PR #40, opened 2019-11-06
Add one more example for route dependency

prostomarkeloff:


frankie567:
Hi @prostomarkeloff! Thank you for your contribution, very appreciated 🙂That's an interesting addition to the docs. I'll make some comments 😉

docs/usage/dependency-callables.md
```diff
@@ -34,3 +34,12 @@ Get the current superuser. Will throw a `401 Unauthorized` if missing or wrong c
 def protected_route(user: User = Depends(fastapi_users.get_current_superuser)):
     return f'Hello, {user.email}'
 ```
+
```
frankie567: Could you add a subsection (`##`) named "In path operation"? Could be interesting also to link the related page in FastAPI documentation: https://fastapi.tiangolo.com/tutorial/dependencies/dependencies-in-path-operation-decorators/

docs/usage/dependency-callables.md
```diff
@@ -34,3 +34,12 @@ Get the current superuser. Will throw a `401 Unauthorized` if missing or wrong c
...
+Or, if you don't need a user, you can use more clear way:
+
+```py
+@app.get('/protected-route', dependencies=[Depends(fastapi_users.get_current_superuser)])
+def protected_route():
+    # do something here...
```
frankie567: Could you remove the comment for consistency with above examples?

PR #47, opened 2019-11-17
Fix #42: multiple authentication backends

### LINKED ISSUE #42: Improve authentication
Currently, only one authentication backend is allowed at a time. This is wrong. One could wish for example to use both Cookie and a Bearer token strategies.

Still taking inspiration from Django, the goal would be to be able to define a list of auth backends. There would be then a wrapper that would run sequentially until it finds a strategy that yield a user ; or raise 401.

Here is a non-working first attempt: 5160dba0043dadc57c36093573d44e7d852e31c7

This approach is promising but currently blocked by https://github.com/tiangolo/fastapi/issues/679.

frankie567:
Fix for #42 
 
Still blocked by https://github.com/tiangolo/fastapi/issues/679 for now ; but the implementation is almost complete.

Remaining tasks:

- [x] Update docstrings
- [x] Update docs
- [ ] Check if interactive docs are OK
- [x] Implement a Cookie authentication as a proof of concept?

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Add quickly a registration and authentication system to your [FastAPI](https://f
 * [X] Customizable database backend
     * [X] SQLAlchemy async backend included thanks to [encode/databases](https://www.encode.io/databases/)
     * [X] MongoDB async backend included thanks to [mongodb/motor](https://github.com/mongodb/motor)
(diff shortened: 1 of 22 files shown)


prostomarkeloff:
In fastapi v0.44 issue has already merged and we can work on this :fire:

PR #97, opened 2020-02-04
Fix tortoise import error

rnd42:
I was getting an import error using tortoise 0.1.1: `ImportError: cannot import name 'Model' from 'tortoise'`.  These few changes seemed to have resolved the issue.


frankie567:
It does seem indeed to match what Tortoise recommend in their docs. Thank you very much 👍

PR #98, opened 2020-02-04
Update pyproject.toml to use passlib 1.7.2

rnd42:
I was getting DepricationWarnings from using passlib 1.7.1.  It seems like Pipfile has already been updated for 1.7.2 but pyproject.toml got left behind.


frankie567:
Nice catch, thanks 👍I try to think about it whenever Dependabot makes a PR, but it seems I missed this one 🙂

frankie567:
@all-contributors add @rnd42 for code

PR #210, opened 2020-05-31
fix documentation tokenUrl

nullhack:
**Problem**

Swagger documentation of [full example files](https://frankie567.github.io/fastapi-users/configuration/full_example/) cannot authorize token to be used, because tokenUrl is not set.

**Solution**

[Change the default tokenUrl](https://github.com/frankie567/fastapi-users/blob/master/fastapi_users/authentication/jwt.py#L33) of `JWTAuthentication` methods of all `example files`.

Issue (https://github.com/frankie567/fastapi-users/issues/211)

### DIFF
--- a/docs/src/full_mongodb.py
+++ b/docs/src/full_mongodb.py
@@ -40,7 +40,7 @@ def on_after_forgot_password(user: UserDB, token: str, request: Request):
     print(f"User {user.id} has forgot their password. Reset token: {token}")
 
 
-jwt_authentication = JWTAuthentication(secret=SECRET, lifetime_seconds=3600)
+jwt_authentication = JWTAuthentication(secret=SECRET, lifetime_seconds=3600, tokenUrl='/auth/jwt/login')
 
 app = FastAPI()
 fastapi_users = FastAPIUsers(
--- a/docs/src/full_sqlalchemy.py
+++ b/docs/src/full_sqlalchemy.py
@@ -51,7 +51,7 @@ def on_after_forgot_password(user: UserDB, token: str, request: Request):
     print(f"User {user.id} has forgot their password. Reset token: {token}")
 
 
-jwt_authentication = JWTAuthentication(secret=SECRET, lifetime_seconds=3600)
+jwt_authentication = JWTAuthentication(secret=SECRET, lifetime_seconds=3600, tokenUrl='/auth/jwt/login')
 
 app = FastAPI()
 fastapi_users = FastAPIUsers(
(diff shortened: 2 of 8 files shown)


nullhack:
Result after applying the changes image

frankie567:
Hi @nullhack!

I'm really I couldn't participate in the conversations this weekend. I really appreciate your help. This fix is something I thought to do at one point when coding, and then obviously forgot 😅 Thank you for this, much appreciated!

nullhack:
> Thank you for that! If I may ask, could you make it as default in the `JWTAuthentication` class? Here : https://github.com/frankie567/fastapi-users/blob/master/fastapi_users/authentication/jwt.py#L33

One problem of making `/auth/jwt/login` as default of `JWTAuthentication` class is that this is not a default path for the rest of the API.

The [examples](https://frankie567.github.io/fastapi-users/configuration/full_example/) set prefix like:

```
app.include_router(
    fastapi_users.get_auth_router(jwt_authentication), prefix="/auth/jwt", tags=["auth"]
)
```

The default path (if no prefix is set) is `/login`. So I think this should be the standard default of `JWTAuthentication` (instead of `/users/login`) and in case prefix is set, then either the user needs to set `tokenUrl` himself or another clever way to make `tokenUrl` be automatically set as `prefix` from the router.

martincolladofab:
I'm agree with @nullhack, probably the `JWTAuthentication` object must set the `tokenUrl` inside the `get_auth_router` call (that call is what setup the router before added to the `app.include_router()`

I don't know if it's possible, but possible idea would be when you pass to the `get_auth_router` the JWTAutenthication object, you setup the tokenUrl for the Swagger Docs. Probably the `get_auth_router` could accept an extra parameter `prefix=` that would setup the router prefix and the JWTAuth tokenUrl. That's possible? If you pass to the FastAPI `app.include_router` a router with a prefix, that's what it's taken for the prefix on that router?

martincolladofab:
What action is necessary to merge and finish this PR?

nullhack:
@martincolladofab My change is just on the examples. The suggestion from @frankie567 implies changes on the standard value. But that suggestion (as I see It) breaks consistency. If done by the back-end, the default value needs to be discussed, so I can change It and ask for a review.

frankie567:
Sorry for not being around lately guys, had quite some other work to attend to. I get your point about the default value @nullhack, and I agree. Let's put `/login` as default then.

I also get your idea @martincolladofab, but for now, this will do.

PR #213, opened 2020-06-01
fix "Authorization: Bearer Undefined" on Swagger

nullhack:
This PR solves the issue https://github.com/frankie567/fastapi-users/issues/212

**Problem**

After authentication on Swagger documentation, the button `Try it out` generates `Undefined` instead of filling the Token of authenticated user.

**Solution**

Change the return of `get_login_response` on [jwt.py](https://github.com/frankie567/fastapi-users/blob/master/fastapi_users/authentication/jwt.py#L68) to match the expected return described at [fastapi's guides](https://github.com/tiangolo/fastapi/blob/master/docs_src/security/tutorial004.py#L128)

**Testing**

image

### DIFF
--- a/tests/test_authentication_jwt.py
+++ b/tests/test_authentication_jwt.py
@@ -68,7 +68,7 @@ class TestAuthenticate:
 async def test_get_login_response(jwt_authentication, user):
     login_response = await jwt_authentication.get_login_response(user, Response())
 
-    assert "token" in login_response
+    assert "access_token" in login_response
 
     token = login_response["token"]
     decoded = jwt.decode(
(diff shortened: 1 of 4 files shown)


frankie567:
Very nice catch, @nullhack! Thank you very much 👍

tests/test_authentication_jwt.py
```diff
@@ -68,9 +68,9 @@ async def test_valid_token(self, jwt_authentication, mock_user_db, token, user):
 async def test_get_login_response(jwt_authentication, user):
     login_response = await jwt_authentication.get_login_response(user, Response())
 
-    assert "token" in login_response
+    assert "access_token" in login_response
```
frankie567, line 71: We should add `assert login_response["token_type"] == "bearer"` here

nullhack: Sure, I'll do during the weekend

martincolladofab:
What action is necessary to merge and finish this PR?

nullhack:
@martincolladofab This is done. Just need a review and merge.

frankie567:
Good to go! Thank you very much @nullhack!

PR #282, opened 2020-07-30
Add database adapter for AWS DynamoDB

### LINKED ISSUE #271: Create a database adaptor for AWS DynamoDB
It can be added using [`boto3`](https://github.com/boto/boto3) or even something higher level such as [`pynamodb`](https://github.com/pynamodb/PynamoDB). What do you think? If you wanna do this I can help you if you need. Or I can take this up myself if you are not familiar with AWS, with some help from you I can do this.

SkullTech:
Add database adapter for AWS DynamoDB. 

Agenda
- [x] Add database adapter
- [ ] Add tests
- [ ] Add documentation

Creating this so that @frankie567 can take a look at it. I'll keep writing the tests and documentation and pushing those commits in the meantime.

A successful merge will close #271 .

### DIFF
--- a/fastapi_users/db/dynamodb.py
+++ b/fastapi_users/db/dynamodb.py
@@ -1,0 +1,160 @@
+import json
+from typing import Mapping, Optional, Type
+
+import aioboto3
+from boto3.dynamodb.conditions import Key
+from pydantic import UUID4
+
+from fastapi_users.db.base import BaseUserDatabase
+from fastapi_users.models import UD
+
+
+class NotSetOAuthAccountTableError(Exception):
+    """
+    OAuth table was not set in DB adapter but was needed.
+
+    Raised when trying to create/update a user with OAuth accounts set
+    but no table were specified in the DB adapter.
+    """
+
+    pass
+
+
+class DynamoDBUserDatabase(BaseUserDatabase[UD]):
+    """
+    Database adapter for AWS DynamoDB.
+
+    :param user_db_model: Pydantic model of a DB representation of a user.
+    :param collection: Collection instance from `motor`.
+    """
+
+    def __init__(
+        self,
+        user_db_model: Type[UD],
+        resource_args: Mapping,
+        users_table: str,
+        users_email_index: str,
+        oauth_accounts_table: Optional[str] = None,
+        oauth_accounts_user_index: Optional[str] = None,
+        oauth_accounts_account_index: Optional[str] = None
+    ):
+        super().__init__(user_db_model)
+        self.resource_args = resource_args
+        self.users_table = users_table
+        self.users_email_index = users_email_index
+        self.oauth_accounts_table = oauth_accounts_table
+
+        if self.oauth_accounts_table and not (oauth_accounts_user_index and oauth_accounts_account_index):
+            raise ValueError("index names must be provided for oauth accounts table")
+        self.oauth_user_index = oauth_accounts_user_index
+        self.oauth_account_index = oauth_accounts_account_index
+
+    async def get(self, id: UUID4) -> Optional[UD]:
+        async with aioboto3.resource("dynamodb", **self.resource_args) as db:
+            table = await db.Table(self.users_table)
+            response = await table.get_item(Key={"id": str(id)})
+            user = response.get("Item")
+            return await self._make_user(user) if user else None
+
+    async def get_by_email(self, email: str) -> Optional[UD]:
+        async with aioboto3.resource("dynamodb", **self.resource_args) as db:
+            table = await db.Table(self.users_table)
+            response = await table.query(
+                IndexName=self.users_email_index,
+                KeyConditionExpression=Key("email").eq(email)
+            )
+            results = response.get("Items")
+            user = results[0] if results else None
+        return await self._make_user(user) if user else None
+
+    async def get_by_oauth_account(self, oauth: str, account_id: str) -> Optional[UD]:
+        if self.oauth_accounts_table:
+            async with aioboto3.resource("dynamodb", **self.resource_args) as db:
+                table = await db.Table(self.oauth_accounts_table)
+                response = await table.query(
+                    IndexName=self.oauth_account_index,
+                    KeyConditionExpression=Key("oauth_name").eq(oauth) & Key("account_id").eq(account_id)
+                )
+                results = response.get("Items")
+                if results:
+                    oauth_account = results[0]
+                    user = await self.get(oauth_account.user_id)
+                    return await self._make_user(user.dict()) if user else None
+        raise NotSetOAuthAccountTableError()
+
+    async def create(self, user: UD) -> UD:
+        user_dict = json.loads(user.json())
+        oauth_accounts_values = None
+
+        if "oauth_accounts" in user_dict:
+            oauth_accounts_values = []
+
+            oauth_accounts = user_dict.pop("oauth_accounts")
+            for oauth_account in oauth_accounts:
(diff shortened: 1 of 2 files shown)


fastapi_users/db/dynamodb.py
```diff
@@ -0,0 +1,160 @@
...
+                    for oauth_account in oauth_accounts_values:
+                        await batch.put_item(Item=oauth_account)
+
+            table = await db.Table(self.users_table)
+            await table.put_item(Item=user_dict)
+
+        return user
+
+    async def delete(self, user: UD) -> None:
+        user_dict = json.loads(user.json())
```
frankie567, line 142: Instead of generating and reparsing JSON, you could directly do `user.dict()`.

fastapi_users/db/dynamodb.py
```diff
@@ -0,0 +1,160 @@
...
+        oauth_accounts_user_index: Optional[str] = None,
+        oauth_accounts_account_index: Optional[str] = None
+    ):
+        super().__init__(user_db_model)
+        self.resource_args = resource_args
+        self.users_table = users_table
+        self.users_email_index = users_email_index
+        self.oauth_accounts_table = oauth_accounts_table
+
+        if self.oauth_accounts_table and not (oauth_accounts_user_index and oauth_accounts_account_index):
```
frankie567, line 47: Unless I'm mistaken, DynamoDB is a NoSQL database, so it may be easier to implement OAuth accounts as sub-documents, like we do in the MongoDB adapter. What do you think?

fastapi_users/db/dynamodb.py
```diff
@@ -0,0 +1,160 @@
...
+
+    pass
+
+
+class DynamoDBUserDatabase(BaseUserDatabase[UD]):
+    """
+    Database adapter for AWS DynamoDB.
+
+    :param user_db_model: Pydantic model of a DB representation of a user.
+    :param collection: Collection instance from `motor`.
```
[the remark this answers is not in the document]
SkullTech: Yes of course, I'll do that, I missed it. Thanks.

fastapi_users/db/dynamodb.py
```diff
@@ -0,0 +1,160 @@
...
+                )
+                results = response.get("Items")
+                if results:
+                    oauth_account = results[0]
+                    user = await self.get(oauth_account.user_id)
+                    return await self._make_user(user.dict()) if user else None
+        raise NotSetOAuthAccountTableError()
+
+    async def create(self, user: UD) -> UD:
+        user_dict = json.loads(user.json())
```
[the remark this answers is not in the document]
SkullTech: @frankie567 Yeah, about that. I was using `user.dict()` at first, but that doesn't take care of GUID the way I want. I want the serialized `dict` to contain a string representation of GUID, just using `user.dict()` is not enough for that.

So to tackle that, I first wrote a helper method ` _serialize(self, user)` that takes an user object as input, does `user.dict()` on it, and replaces the GUID with a string representation using `id.hex` or `str(id)`, either works. But that seemed sloppy to me, so I though `json.loads(user.json())` would be more elegant, efficiency-wise also both approach should be similar.

So yeah, how do you suggest should we tackle this? I need a serialized version of the user with a string id for dynamodb. If you have a more elegant method we can go with that. Or if you prefer writing a custom method like `_serialize` we can do that too.

fastapi_users/db/dynamodb.py
SkullTech: Well DynamoDB is more of a key-value store than document-based store, a bit different than Mongo or Elasticsearch. As a result, it doesn't support queries such as the following used the Mongo adapter

```python
user = await self.collection.find_one(
    {
        "oauth_accounts.oauth_name": oauth,
        "oauth_accounts.account_id": account_id,
    }
)
```
It doesn't support querying using attributes of dicts inside lists. I've looked around and I'm pretty sure of that, see [this](https://stackoverflow.com/questions/42505089/dynamodb-how-to-store-a-list-of-items), [this](https://stackoverflow.com/questions/44990872/dynamodb-query-by-contains-one-of-the-values-from-array-for-node-js), [this](https://stackoverflow.com/questions/34211436/find-all-items-where-a-list-field-contains-a-value-in-dynamodb), [this](https://stackoverflow.com/questions/26980127/filtering-querying-by-the-contents-of-a-list-in-dynamodb), [this](https://stackoverflow.com/questions/50393081/how-can-i-query-list-of-objects-in-dynamodb-using-cli), and [this](https://stackoverflow.com/questions/56547138/scan-dynamo-db-with-boto3-for-array-of-dictionary) stack overflow answers for reference.

As far as I understand, it can be taken care of in one of the following ways

- For every user, keep a list of strings in the format `"oauth_account:account_id"`, the `:` is just acting as a separator. Having this list of strings, DynamoDB can execute a `CONTAINS` condition that will return only the users having a specific string in that list. This way, we can implement `get_by_oauth_account`. The downside is that, `CONTAINS` is only supported in a `SCAN` operation, it isn't supported in a `QUERY` operation. SCAN is an inefficient operation which looks at all the records in a table, whereas `QUERY` is efficient and uses an index. So, using `SCAN` operations will increase the dynamodb costs. 
- Just use an SQL-like apporach like I've done. This keeps the read costs to minimum as we can use `QUERY` for all the calls, but it requires the client to make multiple API calls.
- A third approach would be duplicating the user entries for every oauth account, for example, if there are 3 oauth accounts for a single user, there would be three entries in the database corresponding to that user. This seems a bit ridiculous, as we're using wasting storage space. But it might make sense in the context of dynamodb where read and writes are costly, but storage is inexpensive. If we go with this approach, then we can use single `QUERY` operation API calls for all the methods, so it's efficient in that way. But it takes more storage space.

I'm not an expert on DynamoDB, but this is my understanding so far. I'll keep researching and might ask some experts on Twitter too. @frankie567 If you're not that familiar with DynamoDB you can read [this article](https://www.dynamodbguide.com/mongo-db-vs-dynamo-db), very helpful for understandin the core differences between MongoDB and DynamoDB.

SkullTech: @frankie567 Just read [this](https://www.alexdebrie.com/posts/dynamodb-one-to-many/) article which lays out different strategies to model one-to-many relationships in DynamoDB. Seems like [the duplication strategy](https://www.alexdebrie.com/posts/dynamodb-one-to-many/#denormalization-by-duplicating-data) is the most appropriate for our use case. I'm looking into it more, I'll keep you updated.

SkullTech: I talked with Alex Debrie on Twitter about this, and he agreed that given the `get_by_oauth_account` method doesn't get called that frequently, it would be best to go with the SQL-like approach I've taken. The other solution, i.e. duplication would bring a lot of headaches around consistency and the performance benefits won't be that much either.

frankie567: Hi! Thank you for all this material. I agree by the light of this that the approach you took is the way to go then.

fastapi_users/db/dynamodb.py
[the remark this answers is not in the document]
frankie567: What bugs me is that we serialize it into a string, to parse it again just after ; it seems to be a waste of resources. I don't mind having an helper method like this:

```py
def dict_str_id(self):
    return {**self.dict(), "id": str(self.id)}
```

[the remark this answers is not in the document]
SkullTech: Yeah I get you, makes sense. Will do that.

Issue #343, opened 2020-09-27
VARCHAR requires a length on dialect mysql


lill74:
Hello
# Problem
There is something error on creating tables on MySQL.
`
sqlalchemy.exc.CompileError: (in table 'user', column 'email'): VARCHAR requires a length on dialect mysql
`
effected columns
SQLAlchemyBaseUserTable:
email
hashed_password

SQLAlchemyBaseOAuthAccountTable:
oauth_name
access_token
refresh_token
account_id
account_email

https://github.com/frankie567/fastapi-users/blob/0134a43fade100c32588a0bf09d21b1939d24c8e/fastapi_users/db/sqlalchemy.py#L56

According to SQLAlchemy Docs, 
`CREATE TABLE DDL is issued if a VARCHAR with no length is included.`

# Solution
Pull request #344

lill74:
I think the best way of solve this problem is adding length to columns.

nicolaipre:
I am experiencing the same problem.

PR #344, opened 2020-09-27
Update sqlalchemy.py

lill74:
Solution for Issue: #343 

# The reason why i set length like this
Email address can't exceed 320 References: https://www.rfc-editor.org/rfc/rfc3696.txt
Hashed Password(on bcrypt)'s maximum length is 72
Tokens are must to have sufficient length

### DIFF
--- a/fastapi_users/db/sqlalchemy.py
+++ b/fastapi_users/db/sqlalchemy.py
@@ -53,8 +53,8 @@ class SQLAlchemyBaseUserTable:
     __tablename__ = "user"
 
     id = Column(GUID, primary_key=True)
-    email = Column(String, unique=True, index=True, nullable=False)
-    hashed_password = Column(String, nullable=False)
+    email = Column(String(length=320), unique=True, index=True, nullable=False)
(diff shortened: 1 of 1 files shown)


frankie567:
Sounds sensible, thanks! However, I'm a bit surprised at the length you provided. I assumed VARCHAR couldn't exceed 255 characters. Am I wrong on this?

lill74:
VARCHAR's length can be up to 65,535 after MySQL 5.0.3.
[11.3.2 The CHAR and VARCHAR Types](https://dev.mysql.com/doc/refman/5.7/en/char.html)
mysql_8_adoption_rate-768x450

And up to 85% uses a newer version than MySQL 5.1.
[MySQL 8.x Adoption Rate – Survey Results](https://www.eversql.com/mysql-8-adoption-usage-rate/)

frankie567:
How, since 2005! Definitely, my DB teacher was not up to date on this (I passed my CS degree in 2013!). Thank you!

All good, then!

frankie567:
@all-contributors add @lill74 for bug, code

PR #402, opened 2020-11-30
Use `Protocol` from `typing` if available

roywes:
Some of the 'experimental' definitions in typing_extensions are added to the typing module in time.

This PR prefers the stdlib imports when available, falling back to typing_extensions as necessary


frankie567:
Good catch! Thanks 👍

frankie567:
@all-contributors add @roywes for code

PR #439, opened 2021-01-09
Finalize user activation feature

frankie567:
Just some last tweaks and improvements before merging it in master. All credits go to @mark-todd and @eddsalkield.


frankie567:
I think I'm quite happy with the code now 😄 I'll start to write the documentation!

PR #470, opened 2021-02-05
Ormar backend support

paolodina:
@frankie567 

Hello, this PR adds the support for [ormar](https://collerek.github.io/ormar/) db backend. Reference ticket: https://github.com/collerek/ormar/issues/91

This is briefly what has been done (with the help of ormar maintainer, kudos):
* create the ormar adapter
* add tests
* update docs
* update dependecies files

A couple of notes:
- I see a lot of mypy errors in VS Code but I don't know how to fix
- about the fork, I'm not sure I did a good job with the latest two commits, should I do some change to the repo? (squash or somethings else)

If you have a look at this PR, please let me know! Thanks


paolodina:
I run `make test` locally but I didn't notice the issue. I'll try to fix.

collerek:
If you want any help I can assist ;) Let me know.

frankie567:
Thank you for this, looks good! I'll review it thoroughly 😀

fastapi_users/db/ormar.py
```diff
@@ -0,0 +1,128 @@
...
+        oauth_accounts = getattr(user, "oauth_accounts", [])
+        model = await self.model(**user.dict(exclude={"oauth_accounts"})).save()
+        if oauth_accounts and self.oauth_account_model:
+            await self.create_oauth_models(model=model, oauth_accounts=oauth_accounts)
+        return user
+
+    async def update(self, user: UD) -> UD:
+        oauth_accounts = getattr(user, "oauth_accounts", [])
+        model = await self.get_db_user(id=user.id)
+        # have no idea why other backends does not check if user exists?
```
frankie567, line 115: What happens in other adapters is that we directly make a request with the ORM. So, if the user does not exist, the ORM exception is directly raised. 

What I propose is that you don't swallow the exception in `get_user_db`, but rather in `get_user`:

```py

    async def get_db_user(self, **kwargs: Any) -> Optional[ormar.Model]:
        query = self.model.objects.filter(**kwargs)
        if self.oauth_account_model is not None:
            # in ormar you can select from reverse too
            # only one query is run - faster execution
            query = query.select_related("oauth_accounts").order_by(
                "oauth_accounts__created_date"
            )
        return await query.get()

    async def get_user(self, **kwargs: Any) -> Optional[UD]:
        try:
            user = await self.get_db_user(**kwargs)
            return self.user_db_model(**user.dict())
        except NoMatch:
            return None
```

This way, you won't have to check it here.

fastapi_users/db/ormar.py
```diff
@@ -0,0 +1,128 @@
...
+        # is it some pattern that we have no exception at all?
+        if not model:
+            raise NoMatch("User with given id does not exist!")
+        await model.update(**user.dict(exclude={"oauth_accounts"}))
+
+        if oauth_accounts and self.oauth_account_model:
+            # we issued query with select_related so we have oauths if they exist
```
frankie567, line 122: Ok! You can remove this comment.

docs/src/full_ormar.py
```diff
@@ -0,0 +1,105 @@
+import databases
```
frankie567, line 1: Don't forget to add this example in `full_example.md`

frankie567 (changes requested):
Great work :) Some small changes and we'll be good to go!

fastapi_users/db/ormar.py
```diff
@@ -0,0 +1,128 @@
...
+
+    id = ormar.UUID(primary_key=True, uuid_format="string")
+    oauth_name = ormar.String(nullable=False, max_length=255)
+    access_token = ormar.String(nullable=False, max_length=255)
+    expires_at = ormar.Integer(nullable=True)
+    refresh_token = ormar.String(nullable=True, max_length=255)
+    account_id = ormar.String(index=True, nullable=False, max_length=255)
+    account_email = ormar.String(nullable=False, max_length=255)
+    # added to keep ordering by creation_date
+    created_date = ormar.DateTime(default=datetime.datetime.now)
```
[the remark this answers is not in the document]
collerek: Since some backends (i.e. postgress) do not store the rows in order `ormar` always issues a sort (order_by) even if you do not provide one explicitly. By default it's a primary_key column and since you use `uuids` the order can be pseudo random.

I didn't know if you care to return them in order assumed so by looking at this test that is copied from other db adapters (tortoise etc.).
```python
 # Update
    user_db.oauth_accounts[0].access_token = "NEW_TOKEN"
    await ormar_user_db_oauth.update(user_db)

    # Get by id
    id_user = await ormar_user_db_oauth.get(user.id)
    assert id_user is not None
    assert id_user.id == user_db.id
    # if you do not care about the order here you need to check by id of oauth as you
    # are not guaranteed to get the same oauth in same place without explicit ordering after update
    # since most of the backends (incuding ormar) deletes and recreates the oauths
    # i.e. to populate the default values on model set as server_defaults -> you need to refetch the oauths from db
    # and not just return the object you pass to update
    assert id_user.oauth_accounts[0].access_token == "NEW_TOKEN"
```

If you do not care this column can be removed and the test changed to find oauth by id and not by index in a list.

fastapi_users/db/ormar.py
collerek: That is not true for all adapters.
Simply add this part for sqlalchemy backend tests:

```python
    # Update
    user_db.is_superuser = False # change to False
    await sqlalchemy_user_db.update(user_db)

    # NEW - fail update not fails
    user_db.is_superuser = True
    old_id = user_db.id
    user_db.id = uuid.uuid4()  # new id that does not exist
    await sqlalchemy_user_db.update(user_db)  # should fail but runs query that changes nothing
    user_db.id = old_id 
    # END OF NEW   
    
    # since we not updated anything the old one fails
    # Get by id
    id_user = await sqlalchemy_user_db.get(user.id)
    assert id_user is not None
    assert id_user.id == user_db.id
    assert id_user.is_superuser is True # fails here as there was no update
```

Didn't check the other adapters to be honest.

fastapi_users/db/ormar.py
```diff
@@ -0,0 +1,128 @@
...
+from fastapi_users.db.base import BaseUserDatabase
+from fastapi_users.models import UD, BaseOAuthAccount
+
+
+class OrmarBaseUserModel(ormar.Model):
+    class Meta:
+        tablename = "users"
+        abstract = True
+
+    # sqlalchemy adapter uses char(36) so for ormar it's string format
```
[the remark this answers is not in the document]
collerek: Yeah the comments were intended for @paolodina ;)

collerek:
@paolodina I created a PR to your branch with requested changes, feel free to check it out.

fastapi_users/db/ormar.py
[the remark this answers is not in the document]
frankie567: Ok, I see. I can confirm you that the order doesn't matter. The fact that I'm using `[0]` index in the unit test is more a lazy/convenient shortcut than a real feature requirement 😄

You can remove the `created_date` field then ; as for the test, I'll revisit this later.

fastapi_users/db/ormar.py
frankie567: Hmm, maybe there should be a check and a exception in the `update` method, indeed. But this is out of the scope of this PR.

Yes, I confirm that I expect `None` to be returned by `get`, `get_by_email` and `get_by_oauth_account` method. However, you can very well raise it from your private method `get_db_user` ; to ensure an exception is raised if we try to update an unknown user.

As for reinstantiating back the user from the DB response, my assumption is that Pydantic model and DB model should be equivalent. So database should not do anything but writing the data in the Pydantic model.

collerek: Yes I agree about the exception but if we move the exception check to _get_user it will work fine without additional check in update. But in this version (before my pr to @paolodina fork) `ormar` did not refresh the fields from db, just returned passed user form pydantic (tortoise is doing the same)

What if you have a nullable column that have a server_default setting (possible in `ormar`, in some other backends inqluding `sqlalchemy` too). For example updated_date that should use the sql server datetime and not the python app one.

This field is null when passed from fastapi-users (and can be as it's nullable, no default value in pydantic too) and will be populated only during the query. Or another example is just onupdate "trigger" that always updates the update date but when the passed to db adapter pydantic model is null or the previous one. Without refetch/parsing the db model the returned original model would be outdated.

in example new version for ormar is (to be merged from pr):
```python
    async def create(self, user: UD) -> UD:
        oauth_accounts = getattr(user, "oauth_accounts", [])
        model = await self.model(**user.dict(exclude={"oauth_accounts"})).save()
        if oauth_accounts and self.oauth_account_model:
            await self._create_oauth_models(model=model, oauth_accounts=oauth_accounts)
        user = await self._get_user(id=user.id)  # <- refresh the user from db - make sure it's all updated
        return user
``` 

and was similiar to tortoise (in my opinion it's wrong):
```python
    async def create(self, user: UD) -> UD: # <- user to save is passed
        user_dict = user.dict()
        oauth_accounts = user_dict.pop("oauth_accounts", None)

        model = self.model(**user_dict)  # < - new model is created and saved
        await model.save()

        if oauth_accounts and self.oauth_account_model:
            oauth_account_objects = []
            for oauth_account in oauth_accounts:
                oauth_account_objects.append(
                    self.oauth_account_model(user=model, **oauth_account) # <- same model is passed to oauths
                )
            await self.oauth_account_model.bulk_create(oauth_account_objects)

        return user # < - but only original user that was passed is returned - that does not have to be the same as model and model is never parsed or refetched to user that is returned

```

paolodina:
Well, if I didn't screw up anything in the process (or resolving the merge conflict in `Pipfile.lock`), the requested changes should be ready.

frankie567 (approved):
All good, then 👍 Thank you guys!

frankie567:
@all-contributors add @collerek  for code

PR #533, opened 2021-03-01
fix textual description to match the code

augusto-herrmann:
I think you mean `requires_verification`, like in the code.


frankie567:
Yes, you're right! Thanks 🙏

@all-contributors add @augusto-herrmann for docs

PR #619, opened 2021-05-01
Updated docs to support pip install via zsh

vladhoi:


frankie567:
Nice catch, using `zsh` myself, that's something I often forget 🙃

However, in order to keep the documentation lightweight and understandable, I would prefer not doing a special case for every shell out there. I've tested with classic `bash`, and it works perfectly with the quotes as well.

So, what I propose is to put the quotes in the existing examples and not have a special warning for `zsh`. Wouldn't you mind doing this? 😄

frankie567:
@all-contributors add @vladhoi for documentation

vladhoi:
@frankie567 thanks for the review! I have updated the pr based on your comments

frankie567:
Perfect! Thanks 👍

PR #636, opened 2021-05-17
Fix #630: relative tokenUrl

### LINKED ISSUE #630: Use relative `tokenUrl` parameter for JWTAuthentication (and docs)
Currently the [JWTAuthentication docs page](https://frankie567.github.io/fastapi-users/configuration/authentication/jwt/) doesn't document the `tokenUrl` parameter, although it does document all its other parameters.

When/if this gets added, it would be worth mentioning that the `tokenUrl` must be relative (i.e. no prefixing '/') if a custom `root_path` is being used within the FastAPI app.  This is because, if an absolute `tokenUrl` is used instead, then the URL of the token route will be relative to the `root_path`, but the URL in the OpenAPI spec (which is derived from `tokenUrl`) will instead be relative to the base of the URL.  Then any front-end which parses the OpenAPI spec for authentication (e.g. Swagger) won't be able to find the correct URL.

This is currently documented [in the FastAPI docs](https://fastapi.tiangolo.com/tutorial/security/first-steps/#fastapis-oauth2passwordbearer) (see first tip), but would be worth mentioning in the FastAPI-Users docs as well.

Also, I think it would be worth changing the default value of `tokenUrl` from `"/token"` to `"token"`, because the relative URL works both without and with a custom `root_path`.  In addition, the [full examples](https://frankie567.github.io/fastapi-users/configuration/full_example/) could be changed to reflect this too.

frankie567:


docs/configuration/authentication/jwt.md
```diff
@@ -11,12 +11,16 @@ SECRET = "SECRET"
 
 auth_backends = []
 
-jwt_authentication = JWTAuthentication(secret=SECRET, lifetime_seconds=3600)
+jwt_authentication = JWTAuthentication(secret=SECRET, lifetime_seconds=3600, tokenUrl="auth/jwt/login")
```
eddsalkield, line 14: Shouldn't the default parameter here be `tokenUrl="login"`, in accordance with the class definition?

docs/configuration/authentication/jwt.md
```diff
@@ -11,12 +11,16 @@ SECRET = "SECRET"
...
 
 auth_backends.append(jwt_authentication)
 ```
 
-As you can see, instantiation is quite simple. You just have to define a constant `SECRET` which is used to encode the token and the lifetime of token (in seconds).
+As you can see, instantiation is quite simple. You have to define three arguments:
+
+* A constant `SECRET` which is used to encode the token.
+* The lifetime of the token in seconds.
+* The exact path of your login endpoint. It'll allow the interactive documentation to automatically discover it and get a working *Authorize* button. It's worth to note that this path should be **relative**, not absolute. You can read more details about this in the [FastAPI documentation](https://fastapi.tiangolo.com/tutorial/security/first-steps/#fastapis-oauth2passwordbearer).
```
eddsalkield, line 23: Perhaps specify that it should _probably_ be relative - there may be a few small use cases where people explicitly want an absolute path.

docs/configuration/authentication/jwt.md
```diff
@@ -11,12 +11,16 @@ SECRET = "SECRET"
...
 auth_backends = []
 
-jwt_authentication = JWTAuthentication(secret=SECRET, lifetime_seconds=3600)
+jwt_authentication = JWTAuthentication(secret=SECRET, lifetime_seconds=3600, tokenUrl="auth/jwt/login")
 
 auth_backends.append(jwt_authentication)
 ```
 
-As you can see, instantiation is quite simple. You just have to define a constant `SECRET` which is used to encode the token and the lifetime of token (in seconds).
+As you can see, instantiation is quite simple. You have to define three arguments:
```
eddsalkield, line 19: You don't have to define three arguments - `secret` and `lifetime_seconds` are the only required arguments.  `tokenUrl` and `name` are optional.  Perhaps it's also worth mentioning the arguments by their name as well?

eddsalkield (commented):
Overall looking good- thanks for taking this on! I've left a few minor comment in the review :)

docs/configuration/authentication/jwt.md
frankie567: I think it's okay to have `auth/jwt/login` because it's the working value when configuring the JWT authentication router when following the documentation. If the user follows this, it should have a working example out-of-the-box (like full examples).

frankie567:
Thank you for the very relevant comments, much appreciated 🙂 I've improved things a bit!

docs/configuration/authentication/jwt.md
eddsalkield: I think this can make sense, since you've specified that the `tokenUrl` "defaults to `login`".

eddsalkield:
Nice improvements - thanks!  I think this is pretty near good to go.  The only other thing I thought of was quoting the strings when showing the defaults, e.g.:

> A constant `SECRET`

becomes

> A constant `"SECRET"`

or

> A constant "SECRET"

Although this is really minor, so feel free to pick whichever (that said, it's probably worth making this consistent throughout the docs.  While looking to see what is done elsewhere, it seems from the [cookie page](https://github.com/frankie567/fastapi-users/blob/dc67ebe23920070a151a01ee9f308634e258742a/docs/configuration/authentication/cookie.md) that it's a mixture of back-ticked and single-quoted).

docs/configuration/authentication/jwt.md
frankie567: Maybe I should change the default actually 🤔

frankie567:
I've improved the consistency accross cookie/jwt documentation. Thank you for your help 👍

PR #638, opened 2021-06-03
Add an option to provide custom router class

ShipraShalini:


frankie567:
Hi @ShipraShalini! Thank you for your contribution!

I get what you did here, but I don't really see in which cases you need a custom router class. Could you elaborate a bit on this?

Cheers 😄

frankie567:
I see! Sorry but, in this case, isn't setting the `default_response_class` argument on FastAPI/APIRouter class sufficient to do this: https://fastapi.tiangolo.com/advanced/custom-response/#default-response-class ?

ShipraShalini:
Thank you. I did not know about `default-response-class`.
When I set the `default-response-class`, then swagger login doesn't work.
It would be nice to have an option to pass `response_class` to the `FastAPIUsers` get router methods. Then I can create two `auth_router`s one for frontend and one for swagger.
I can modify the (or create another) PR for this.

frankie567:
If your response class changes the response body, that's normal. Swagger expects a specific format to be able to retrieve token from the login response.

However, I don't think we need an option, as this can be set on router level, even when already created. For example:

```py
standard_auth_router = fastapi_users.get_auth_router(jwt_authentication)

custom_response_auth_router = fastapi_users.get_auth_router(jwt_authentication)
custom_response_auth_router.default_response_class = CustomResponse

app.include_router(standard_auth_router, prefix="/auth-standard/jwt", tags=["auth"])
app.include_router(custom_response_auth_router, prefix="/auth-custom/jwt", tags=["auth"])
```

Issue #646, opened 2021-06-18
python_type is set to str in GUID


mark-todd:
## Background
Modules like pydantic sqlalchemy (https://github.com/tiangolo/pydantic-sqlalchemy) require information about the python_type to construct pydantic models. This information is typically encoded in the types attribute `impl.python_type`. (https://github.com/tiangolo/pydantic-sqlalchemy/blob/master/pydantic_sqlalchemy/main.py#L27)

## Problem
The FastAPI Users GUID type (https://github.com/frankie567/fastapi-users/blob/master/fastapi_users/db/sqlalchemy.py#L15) does not contain the python_type information, and defaults to the plain python string (as it's impl attribute is CHAR).

## Solution
We replace the CHAR used here with one that is identical but also contains this python_type information.
e.g replacing this line (https://github.com/frankie567/fastapi-users/blob/master/fastapi_users/db/sqlalchemy.py#L22) with:
```
from pydantic import UUID4
class NewChar(CHAR):
     python_type = UUID4
impl = NewChar
```

frankie567:
Thank you for raising this point @mark-todd! Looks sensible to me.

Correct me if I'm wrong, but if I look at [those lines](https://github.com/tiangolo/pydantic-sqlalchemy/blob/master/pydantic_sqlalchemy/main.py#L28-L29), it seems that we could directly set the `python_type` property in the `GUID` class:

```py
class GUID(TypeDecorator):  # pragma: no cover
    """Platform-independent GUID type.
    Uses PostgreSQL's UUID type, otherwise uses
    CHAR(36), storing as regular strings.
    """

    impl = CHAR
    python_type = UUID4

    # ...
```

If so, that would be even more straightforward.

mark-todd:
Hi Frankie,

Thanks for getting back to me on this! So CHAR already has the python_type attribute (where it stores the default python type `str` ). Unfortunately since this is the first place pydantic-sqlalchemy looks for python_type, it would never reach the elif line that takes the python_type from the GUID more directly. So I think the only way is to override the python_type within CHAR itself, even though this is less direct.

frankie567:
Indeed! Well, let's go for the custom `impl` then. Would you mind making a PR for this? For the `impl` class, I suggest the name `UUIDChar`.

mark-todd:
Yep sounds good! And yeah of course, I'll try and get it submitted today

PR #648, opened 2021-06-25
Added python_type into GUID

mark-todd:
Added the python_type to the GUID class, by overriding CHAR. This is useful for libraries such pydantic-sqlalchemy where this attribute is used. (https://github.com/tiangolo/pydantic-sqlalchemy/blob/master/pydantic_sqlalchemy/main.py#L27)

This is to solve issue #646


frankie567:
Perfect, nothing to add! Thanks 🙏

PR #650, opened 2021-07-03
Return empty response with 204

jnu:
I'm getting an error running the `delete` route. FastAPI apparently chooses JSONResponse by default which encodes `None` as `null`, which means the 204 response has a body, which is an error.

I would think this should be fixed in FastAPI or Starlette, but for now the solution seems to be to return a `Response` directly to ensure FastAPI doesn't use a JSONResponse: https://github.com/tiangolo/fastapi/issues/717#issuecomment-583826657

### DIFF
--- a/fastapi_users/router/users.py
+++ b/fastapi_users/router/users.py
@@ -1,6 +1,6 @@
 from typing import Any, Callable, Dict, Optional, Type, cast
 
-from fastapi import APIRouter, Depends, HTTPException, Request, status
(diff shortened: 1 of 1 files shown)


frankie567:
Indeed, I already stumbled upon this weird behavior of FastAPI. This should be fixed on the framework clearly, but it sticks for quite some time now.

Let's go for this then! Thank you for your contribution 🙏

frankie567:
@all-contributors add @jnu for bug

PR #661, opened 2021-07-13
Add dummy `setup.py` that reuses our `setup.cfg`

### LINKED ISSUE #660: Provide dummy setup.py
### Discussed in https://github.com/frankie567/fastapi-users/discussions/659

<div type='discussions-op-text'>

<sup>Originally posted by **XaviTorello** July 13, 2021</sup>
This will simplify the installation of this package as a `path: <PATH>, develop = True` `poetry` dependency or enforcing a `pip -e install <PATH>`</div>

XaviTorello:
It provides a dummy `setup.py` that reuses our `setup.cfg`.

This will support
- Poetry `my-package = {path = "<PATH>", develop = true}` dependencies
- `pip install -e <PATH>` dependencies

It also defines `setuptools` as a build dependency and extracts the version from `fastapi_users.__version__` to avoid hardcoded duplicates.

Fix #660

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["flit_core >=2,<3"]
+requires = ["flit_core >=2,<3", "setuptools"]
 build-backend = "flit_core.buildapi"
 
 [tool.flit.metadata]
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,9 @@
+[metadata]
+name = fastapi-users
+version = attr: fastapi_users.__version__
+
 [bumpversion]
-current_version = 6.1.1
+current_version = attr: fastapi_users.__version__
 commit = True
 tag = True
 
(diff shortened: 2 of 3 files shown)


frankie567:
Hi there!

Thank you for your proposal! I'm not sure to understand what you are trying to solve here. Could you elaborate a bit? And give an example of what is not working with Poetry?

XaviTorello:
Hi! Sorry for the delay.

With this we'll provide compatibility with editable mode installations.

When developing or integrating new dependant functionalities between projects that uses fastapi-users it can be useful to install it in editor/develop mode (`pip install -e` / poetry develop flag).

This will permit us to simply patch current (virtual)env without touching nothing on the dependencies definition, providing a quick rollback to the expected env just reinstalling the dependencies.

It's just a dummy change that will support more installation scenarios :)

XaviTorello:
Should this be reopened and merged @frankie567? ty

PR #678, opened 2021-08-07
Added username

mawoka-myblock:
Hello!
I am trying/tried to add usernames. It may be unnecessary, I know, but i need it for my project, so why not create a PR?
## Disclaimer
- I am a Noob with SQL-Stuff, so please don't blame me for the bad code.
- I'll revert all the changes to the README before a merge
- **I am very likely too lazy to write the docs, so help would be really nice ❤️ !**

I would be very happy if you would give me some tips how I could make the code better.


mawoka-myblock:
I really struggle with OAuth...

mawoka-myblock:
Yes, I give up for now! I'm having problems with:

- The the registration to check if the username is unique
- With OAuth at all
- Writing useful tests

Things I did:

- Implemented the database-schemas
- Implemented query-functions
- Wrote tests for the database-queries
- Made older tests work with adding an username-property
- Spent 4 1/2 hours on it

PR #705, opened 2021-08-29
Return 403 instead of 401 when a user is known

daanbeverdam:
Hi @frankie567, 

I would like to propose to return a 403 whenever a user is known but doesn't have enough privileges to access a resource and return a 401 if a user could not be authenticated at all (wrong or no credentials provided). This follows the conventions outlined in [rfc7235](https://datatracker.ietf.org/doc/html/rfc7235#section-3.1) and lets `fastapi-users` play nicely with frontend frameworks such as [vue-auth](https://github.com/websanova/vue-auth) which will log you out on receiving a 401.

Let me know if you agree, so I can also update the tests.

### DIFF
--- a/docs/usage/dependency-callables.md
+++ b/docs/usage/dependency-callables.md
@@ -11,7 +11,7 @@ Return a dependency callable to retrieve currently authenticated user, passing t
 
 * `optional`: If `True`, `None` is returned if there is no authenticated user or if it doesn't pass the other requirements. Otherwise, throw `401 Unauthorized`. Defaults to `False`.
 * `active`: If `True`, throw `401 Unauthorized` if the authenticated user is inactive. Defaults to `False`.
-* `verified`: If `True`, throw `401 Unauthorized` if the authenticated user is not verified. Defaults to `False`.
+* `verified`: If `True`, throw `403 Forbidden` if the authenticated user is not verified. Defaults to `False`.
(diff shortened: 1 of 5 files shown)


frankie567:
Hi @daanbeverdam!

Thank you for your proposal!

Actually, I did this on purpose. In my opinion an inactive user should be considered the same as a non-existing user: it shouldn't be allowed to login and existing sessions should expire. This is the [behavior adopted by default in Django](https://docs.djangoproject.com/en/3.2/ref/contrib/auth/#django.contrib.auth.models.User.is_active).

However, concerning the not verified, well, I'm unsure 😅

daanbeverdam:
Ah I see, that makes sense. Then I would propose to return a 401 for inactive users then, and a 403 for the unverified. Because you know who the unverified user is and in many cases want to give them access to parts of the application where verification is not necessary. But if they stumble on a forbidden route we don't want them to logout completely (which is standard behaviour on receiving a 401).

frankie567:
Yes, I think you convinced me there :) Would you mind making the necessary changes and update the tests and documentation?

Cheers!

daanbeverdam:
Thanks @frankie567! I've updated the logic, tests and documentation. Awaiting your feedback.

frankie567:
Perfect! Thank you 🙏

frankie567:
@all-contributors add @daanbeverdam for code

frankie567:
@all-contributors add @daanbeverdam for code

frankie567:
@all-contributors add @daanbeverdam for code

PR #741, opened 2021-09-23
Fix UserManager docs and custom token audience in password reset

maximka1221:
I noticed that the functions in the docs are named incorrectly so I fixed it 😃 
Also, I saw that when using custom a token audience in the password reset feature, the creation of the token still uses the default value instead of the custom one.


frankie567:
Wow, those are very nice catch! Thank you 🙏😄

frankie567:
@all-contributors add @maximka1221 for documentation and bug

PR #767, opened 2021-10-15
Refactor Code Expression

yezz123:
- Replace unused for index with an underscore.
- Use `items()` to directly unpack dictionary values.
- Merge duplicate blocks in conditional.
- Use any() instead of for loop


frankie567:
@all-contributors add @yezz123 for code

PR #768, opened 2021-10-15
Update Documentation to Fix v8.x.x Changes

jdukewich:
Some of the documentation has not been updated to reflect the refactor in v8, especially relating to the new `UserManager`. With this PR, I update the documentation under the `Usage -> Routes` page. For instance, the docs still reference an `after_register` handler, but that is now the `on_after_register` handler within a `UserManager`.

This will address #769


frankie567:
Very nice catch @jdukewich, thank you 🙏

@all-contributors add @jdukewich for docs

Issue #219, opened 2020-06-03
[feature request] scope to limit router usage


nullhack:
Would be nice if routes were limited (or capable of) by some scope.
For example: to register a user, we could set something like:

```
app.include_router(
    fastapi_users.get_register_router(),
    prefix="/auth",
    tags=["auth"],
    scope=["user:register"]
)
```

Then, if the token has the scope or is a superuser, It's allowed to use the router.

This feature depends on https://github.com/frankie567/fastapi-users/issues/218 or else the first user would need to be created directly on database.

frankie567:
I'll make quite the same comment as in #217 : a scope approach could be interesting, but I don't think it's in the scope of this project. I think we can wire things pretty easily between `fastapi-permissions` (for example) and `fastapi-users` with dependency callables.

Issue #217, opened 2020-06-03
[question] Is it possible to avoid database requests using scopes?


nullhack:
I'm trying to use scopes with fastapi-users. Is there a simple example of how to use scopes?

My goal is to use scopes for things that don't require accessing user data (e.g. I don't need to query the database and get user id if my API is just rendering a list (in memory/redis) that requires the user to have read access), for this I can just check the scope.

Is it possible?

martincolladofab:
Hi @nullhack, working on this kind of things this weekend I found this library ([fastapi-permissions](https://github.com/holgi/fastapi-permissions)) and it's used by Netflix on the dispatch project ([dispatch](https://github.com/Netflix/dispatch))
It's not using scopes, but I think that could be a good starting point. @frankie567 maybe could be included in fastapi-users to manage the permissions?

nullhack:
Thank you, I'll take a look.

frankie567:
Currently, the approach is to check against the database if the user id. in the JWT still exists and that it has the right permission (active and superuser).

However, if you implement your own authentication class, I think you can make an authentication that works without DB query. The only constraint is that it returns a valid User object.

I think a scope approach could be interesting, but I don't think it's in the scope of this project. I think we can wire things pretty easily between `fastapi-permissions` (for example) and `fastapi-users` with dependency callables.

nullhack:
The only problem I see is that by design, fastapi-users check against database for every request. This defeats the purpose of permissions from scope (we want to check fast without database requests as much as possible). 

> However, if you implement your own authentication class, I think you can make an authentication that works without DB query

This would be nice.

PR #787, opened 2021-11-03
Use scopes

artreven:
I tried to use scopes inspired by https://fastapi.tiangolo.com/advanced/security/oauth2-scopes/#dependency-tree-and-scopes and discussions in #217 and #219.
The scopes are ingested into the token data, so we can check them without calling the db. However, as I did not want to make changes to existing functionalities, database is still queried on every user request unless the scope check fails.

### DIFF
--- a/fastapi_users/authentication/__init__.py
+++ b/fastapi_users/authentication/__init__.py
@@ -1,6 +1,6 @@
 import re
 from inspect import Parameter, Signature
-from typing import Callable, Optional, Sequence
+from typing import Callable, List, Optional, Sequence
 
 from fastapi import Depends, HTTPException, status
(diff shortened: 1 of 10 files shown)


frankie567:
Hi @artreven!

That's interesting. However, that's not something I want to support right now. I involves too much changes in the data models and may be too specific for the general.

I'm really sorry but I hope you'll understand. In the future, consider opening a [Discussion](https://github.com/fastapi-users/fastapi-users/discussions) where we could talk about this before diving in to code.

artreven:
Thanks for the feedback @frankie567 . No problem at all, I had to implement the scopes anyway, so I just decided to contribute it back. I totally understand your decision.

PR #793, opened 2021-11-08
Updated OAuth2 router generator for better OpenAPI docs

matyasrichter:
This PR:
- adds response_model to the OAuth2 `authorize` route
- changes the `authentication_backend` param's type on the OAuth2 `authorize` route to a dynamically generated enum, so that possible values are shown in generated OpenAPI documentation

There's a breaking change on the `authorize` route which now returns 422 instead of 400 when an incorrect backend name is passed to it. This change is IMO justifiable as pydantic validation is generally preferred in FastAPI.

### DIFF
--- a/docs/usage/routes.md
+++ b/docs/usage/routes.md
@@ -229,9 +229,7 @@ Return the authorization URL for the OAuth service where you should redirect you
     ```
 
 !!! fail "`422 Validation Error`"
-
-!!! fail "`400 Bad Request`"
-    Unknown authentication backend.
+    Invalid parameters - e.g. unknown authentication backend.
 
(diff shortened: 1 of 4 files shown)


frankie567:
Wow, that's *super* clever! I never thought about using a dynamic enum to trigger this validation.

Thank you for this, it's a great contribution 🙏 May I ask you to update the documentation [right here](https://github.com/fastapi-users/fastapi-users/blob/4aef8f8a17f66955ceffd795dc526dbc38bfe0b2/docs/usage/routes.md?plain=1#L233-L234)?

frankie567:
@all-contributors add @matyasrichter for code

PR #798, opened 2021-11-10
Added codestyle check and lint to the build action

matyasrichter:


matyasrichter:
Hm, build failed, odd. I guess the issue is `black` version, then, because it tries to reformat quite a few files when I run it locally.

matyasrichter:
Do you mind if I pin dev dependencies to specific versions @frankie567?

frankie567:
I'll have a look at why this is not working! Thank you for this great addition 👍

matyasrichter:
Must have been an issue on my end :facepalm:

yezz123 (approved):
That looks like a great addon to this package, what do you think about using pre-commit for linting and formatting? 🚀

frankie567:
Thank you @matyasrichter, that's a great addition!

PR #799, opened 2021-11-10
Extending generated OpenAPI docs

matyasrichter:
This is WIP for now, but I'd like to eventually get the generated docs into a state where a frontend dev working with an API that uses fastapi-users wouldn't need to check our documentation. The goal is to have responses included in swagger along with basic descriptions of error codes.


matyasrichter:
Added response examples so that errors are documented in OpenAPI, the output in redoc looks like this:
image
and updated the markdown documentation - some of the responses were outdated.

And I added some basic tests that should (hopefully) prevent inconsistency in the future.



// The python3.7 check for this commit failed due to a network error, could you restart it, please? I think I don't have the required permissions.

matyasrichter:
What do you think about this @frankie567 ?

frankie567:
That's once again a super interesting addition, @matyasrichter! Thank you 🙏
 
May I suggest that `get_login_responses_success` and `get_logout_responses_success` methods on authentication classes be called `get_openapi_login_responses_success` and `get_openapi_logout_responses_success` so that we don't take them for actual API logic?

I see that the codestyle check is failing, maybe just run a `make format` to solve this.

Cheers!

matyasrichter:
@frankie567 fixed :rocket:

matyasrichter:
Hi, sorry to bump this merged PR, but is there an estimate for when it could be released? We'd like to have this ready for our project. @frankie567

frankie567:
Yes, sorry, I'll prepare it!

PR #822, opened 2021-12-16
Removed assignments to a `fastapi_users` variable.

matyasrichter:
On several spots in the docs, the `FastAPIUsers` instance was called `fastapi_users`, potentially leading people to name shadowing. I renamed the variable in those examples to `users`. I also renamed a variable of the same name in tests.

### DIFF
--- a/docs/configuration/routers/auth.md
+++ b/docs/configuration/routers/auth.md
@@ -15,7 +15,7 @@ SECRET = "SECRET"
 
 jwt_authentication = JWTAuthentication(secret=SECRET, lifetime_seconds=3600)
 
-fastapi_users = FastAPIUsers(
+users = FastAPIUsers(
     get_user_manager,
     [jwt_authentication],
     User,
--- a/docs/configuration/routers/index.md
+++ b/docs/configuration/routers/index.md
@@ -17,7 +17,7 @@ Configure `FastAPIUsers` object with all the elements we defined before. More pr
 ```py
 from fastapi_users import FastAPIUsers
 
-fastapi_users = FastAPIUsers(
+users = FastAPIUsers(
     get_user_manager,
     [jwt_authentication],
     User,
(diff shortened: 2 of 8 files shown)


frankie567:
Hi @matyasrichter! TBH, I'm not a big fan of this change. In which context could this be an issue?

matyasrichter:
Here's an example:

First, you create an instance:
```python
# users.py
from fastapi_users import FastAPIUsers

fastapi_users = FastAPIUsers(...)
```
Then, in another file, you import and use that instance - ie. for building a dependency:
```python
# router.py
from users import fastapi_users

current_user = fastapi_users.current_user()

@app.get("/protected-route")
def protected_route(user: User = Depends(current_user)):
    return f"Hello, {user.email}"
```
The issue, in my opinion, is that it's not clear at first glance what `fastapi_users` is. The person reading the code needs to check the imports to see if it's `fastapi_users` - the module, or if it's `fastapi_users` - an instance of a class from the `fastapi_users` module.

Additionaly, if you happen to have a star import somewhere in your app (which I know is kind of bad practice), it's easy to get them mixed up - you can't import the `fastapi_users` library *after* importing the instance, because the instance "overwrites" the name.

Anyway, this PR is nothing major, feel free to close it if you don't agree here. It's just something I've noticed.

frankie567:
I see, but I would like to keep it that way for now. I find it clearer if it's called `fastapi_users`, and I trust developers for keeping their house tidy. In the common case, there is no reason to import the whole `fastapi_users` module on its own. 

Thank you for raising this anyway 🙂

jacksbox:
I'll just comment here, as I run into this very issue, or kind of. I checked the docs on how to access the current user.
But the page in the docs about `current_user` does nowhere state where the `fastapi_users` instance is coming from. Sure thing, after a bit of searching I found it - at the location where I registered the routes for the authentication (as it is in the docs) - by far not the first place I would have expected it.
This is probably no big issue for devs who worked more often with this package. But for a newcomer (like me) - this crucial thing is kinda hidden away. 
Maybe the instanciation should have it's own chapter in the docs (not under routes). And maybe it could be created in an separate file - just to indicate better, that this can/will be used by other parts than just the auth routes?

PR #824, opened 2021-12-17
Fixed #823

### LINKED ISSUE #823: Can't generate OpenAPI schema with two OAuth routers
## Describe the bug

When two OAuth routers are included in one FastAPI app, OpenAPI schema generation fails with KeyError.

## Configuration
 - Python version : 3.9
 - FastAPI version : 0.70.1
 - FastAPI Users version : 8.1.3

matyasrichter:
Closes #823

Basically, FastAPI keeps track of all response models so that they could list them in the generated OpenAPI docs. The dynamic enum from #793 caused an issue with this, because all OAuth routers had a response model of the same name. OpenAPI generation would then crash with a KeyError. 

I added a test for this and switched the OAuth router validation to a regex. It's not as nice for documentation reasons, but I think it's an acceptable solution.

### DIFF
--- a/fastapi_users/authentication/__init__.py
+++ b/fastapi_users/authentication/__init__.py
@@ -1,3 +1,4 @@
+import enum
 import re
 from inspect import Parameter, Signature
 from typing import Callable, Optional, Sequence
@@ -42,6 +43,7 @@ class Authenticator:
     """
 
     backends: Sequence[BaseAuthentication]
+    backends_enum: enum.Enum
 
     def __init__(
         self,
@@ -49,6 +51,10 @@ class Authenticator:
         get_user_manager: UserManagerDependency[models.UC, models.UD],
     ):
         self.backends = backends
+        self.backends_enum = enum.Enum(  # type: ignore
+            "AuthenticationBackendName",
+            {backend.name: backend.name for backend in backends},  # type: ignore
+        )
         self.get_user_manager = get_user_manager
 
     def current_user(
--- a/fastapi_users/router/oauth.py
+++ b/fastapi_users/router/oauth.py
@@ -43,10 +43,6 @@ def get_oauth_router(
             route_name=callback_route_name,
         )
 
-    # matches name of first backend OR name of second backend OR ...
-    auth_backend_names = [backend.name for backend in authenticator.backends]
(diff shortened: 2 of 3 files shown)


frankie567:
Hmm, I see. I'm thinking of two other ways to solve this (not tested it, just random thoughts)

## 1. Instantiate it once in `Authenticator`

Maybe if we create only once in the `Authenticator` class, we avoid the issue ?
 
```py
class Authenticator:

    # ...

    backends: Sequence[BaseAuthentication]

    def __init__(
        self,
        backends: Sequence[BaseAuthentication],
        get_user_manager: UserManagerDependency[models.UC, models.UD],
    ):
        self.backends = backends
        self.get_user_manager = get_user_manager
        self.backends_enum = enum.Enum(
            "AuthenticationBackendName",
            {backend.name: backend.name for backend in authenticator.backends},
        )

    # ...
```

```py
    async def authorize(
        request: Request,
        authentication_backend: authenticator.backends_enum,
        authentication_backend: str = Query(..., regex=auth_backend_regex),
        scopes: List[str] = Query(None),
    ):
```

## 2. Dynamic enum name 

Could it be solved by generating a dynamic enum name. Something like this:

```py
AuthenticationBackendName: enum.EnumMeta = enum.Enum(
    f"AuthenticationBackendName{oauth_client.name}",
    {backend.name: backend.name for backend in authenticator.backends},
)
```

What do you think?

matyasrichter:
Both of your suggestions seem to be working (I mean, tests pass :smile:).

In the first case, mypy complains that `Enum type as attribute is not supported`. But we're not 100% mypy-compliant anyway, are we?

As for the second option, I'm honestly not sure. Again, mypy complains that `Enum() expects a string literal as the first argument`. Other than the API docs, is the name used for anything else? I think Python generates a docstring and maybe a `__str__` and `__repr__` - these shouldn't be an issue, right?

frankie567:
Ok, thank you for checking! My preference would go to the first solution; having the enum on the `Authenticator` class. Indeed, mypy complains when using dynamic variables as type hint, but we already do it plenty of times in the codebase. So another `type: ignore` won't hurt here 😅

matyasrichter:
@frankie567 Updated :+1:

frankie567:
Great! Thanks 🙏

PR #831, opened 2021-12-30
Revamp authentication

frankie567:


fastapi_users/authentication/backend.py
```diff
@@ -0,0 +1,65 @@
...
+        transport: Transport,
+        get_strategy: Callable[..., Strategy[models.UC, models.UD]],
+    ):
+        self.name = name
+        self.transport = transport
+        self.get_strategy = get_strategy
+
+    async def login(
+        self,
+        strategy: Strategy[models.UC, models.UD],
```
davidbrochart, line 43: Is the `strategy` argument needed, or could it be retrieved from `self.get_strategy()`?

frankie567: Not in the general case: `get_strategy` can be a FastAPI dependency (with sub-dependencies) and, in this case, it should be run inside FastAPI dependency injection context.

This is why `get_strategy` is added as dependency in the login route, and passed manually to this login method:

https://github.com/fastapi-users/fastapi-users/blob/96a086c60f99f28c930460b4c180e6853f12183b/fastapi_users/router/auth.py#L46-L69

PR #835, opened 2022-01-03
Allow lifetime_seconds=None

davidbrochart:


frankie567:
Hi @davidbrochart! Thank you for this, that's indeed possible to set `lifetime_seconds` as `None`.

However, it means that the token will be valid **forever**, unless you change the secret. That's okay if you know what you are doing, but it may be a bit dangerous.

Would you mind updating the doc here to add a warning about this:

https://github.com/fastapi-users/fastapi-users/blob/d469bace6eb51e12271705d066c80c498e6b0c7b/docs/configuration/authentication/strategies/jwt.md?plain=1#L19

To:

```md
* `lifetime_seconds` (`Optional[int]`): The lifetime of the token in seconds. Can be set to `None` but in this case the token will be valid **forever**; which may raise serious security concerns.
```

davidbrochart:
Done, thanks! And happy new year :tada:

frankie567:
Thank you! Happy new year, David 😄

frankie567:
@all-contributors add @davidbrochart for code

Issue #846, opened 2022-01-07
Delete response header not setting cookie_samesite as defined in CookieTransport 


Hazedd:
For my angular project i need to set the cookies SameSite header to 'None' so 'Set-Cookie' is set by the angular app. 
I defined the cookie transport:
```
cookie_transport = CookieTransport(cookie_max_age=3600, cookie_samesite='None')
```

On a login call the browser is receiving the following cookie:
> fastapiusersauth=yofyP7VSA3YZ6seMk3YzX075majNRys4vV7N1kO83a0; HttpOnly; Max-Age=3600; Path=/; SameSite=None; Secure

and the cookie will be set.

On a logout call  the browser is receiving the following cookie:
> fastapiusersauth=""; expires=Fri, 07 Jan 2022 11:38:53 GMT; Max-Age=0; Path=/; SameSite=lax

**Setting cookie is blocked by SameSite attribute lax!**

## To Reproduce

setup fastapi-users cookie backend
setup angular project to call login and logout
look in network calls

## Expected behavior
On the delete cookie endpoint return response headers with samesite attribute as defined in the cookie transport.

## Configuration
 - Python version : 3.10
 - FastAPI version : 0.70.1
 - FastAPI Users version : 9.2.0

### FastAPI Users configuration
see [repo](https://github.com/Hazedd/fastapi-users-angular-example) for config.

edit:

I have found the issue:

CookieTransport implements a set and delete cookie of starlette. In the set_cookie of starlette all posible options are present (secure and samesite for this issue). The delete_cookie of starlette makes a expired cookie by using the set_cookie on this point only key, expires max_age path domain are used. 

Not sure deleted cookies should have those parameters, if so bug is in starlette package.

```python
def delete_cookie(self, key: str, path: str = "/", domain: str = None) -> None:
    self.set_cookie(key, expires=0, max_age=0, path=path, domain=domain)
```

fix for fastapi-users:

on CookieTransport class make sure the get_logout_response is using the set_cookie methode of starlette and set cookie-value to ''  and max_age to 0

```python
async def get_logout_response(self, response: Response) -> Any:
    response.set_cookie(
        self.cookie_name,
        '',
        max_age=0,
        path=self.cookie_path,
        domain=self.cookie_domain,
        secure=self.cookie_secure,
        httponly=self.cookie_httponly,
        samesite=self.cookie_samesite,
    )
```

commit 113bcf0 PR: #848

PR #848, opened 2022-01-07
logout response sets proper response headers

Hazedd:
logout response is using starlette delete cookie. In starlette the samesite and secure attributes are not in the header but are needed to set the removed cookie client side. Implementing set_cookie with an empty cookie-value and a max_age of 0 will set a new expired cookie by the client.

related issue #846

### DIFF
--- a/fastapi_users/authentication/transport/cookie.py
+++ b/fastapi_users/authentication/transport/cookie.py
@@ -46,8 +46,15 @@ class CookieTransport(Transport):
         return None
 
     async def get_logout_response(self, response: Response) -> Any:
-        response.delete_cookie(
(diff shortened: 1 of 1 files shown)


frankie567:
Thank you for catching this! Actually, it seems that it has been spotted on encode/starlette#1207 and fixed. However, I think your solution is better for now for backward compatibility with older FastAPI versions.

The CI is failing because of linting issues. Could you run `make format` and commit the fixes?

Hazedd:
> 

sorted now :)

frankie567:
@Hazedd Why closing the PR? I won't be able to merge your fix in this situation 😔

Hazedd:
@frankie567 guess it was a typical sunday.... my bad
see #855

PR #849, opened 2022-01-08
UserTable relation table name correction

Hazedd:
OAuthAccount should be correct table name SQLAlchemyBaseOAuthAccountTable


frankie567:
Good catch, thanks!

PR #891, opened 2022-01-23
chore(ref): Lift code into `else` after break in control flow

yezz123:
Where the body of an if statement ends with a break in the control flow, such as a continue, return or raise, the subsequent statements can be lifted into the else clause.


frankie567:
Hi @yezz123!

Thank you for this proposal! I'm aware of the "ternary" operator in Python; however, I find it less readable in those cases.

If you don't mind, I prefer to keep it that way :)

PR #896, opened 2022-01-24
Generate setup.py file

mekanix:
As Python integration with FreeBSD ports is based on setup.py it would be so much easier to maintain it if that file was generated and uploaded. I am sure that FreeBSD is not the only OS that has this limitation, so potentially this will make porting easier for other operating systems, too.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,7 +32,7 @@ jobs:
         codecov
     - name: Build and install it on system host
       run: |
-        flit build
+        flit build --setup-py
         flit install --deps none --python $(which python)
         python test_build.py
 
@@ -56,4 +56,4 @@ jobs:
         FLIT_USERNAME: ${{ secrets.FLIT_USERNAME }}
         FLIT_PASSWORD: ${{ secrets.FLIT_PASSWORD }}
       run: |
-        flit publish
+        flit publish --setup-py


frankie567:
Hi @mekanix!

Thank you for this proposal! I may not have all the ins and outs about this but, from what I understood, it sounded like `setup.py` was more or less deprecated. There is a long thread about this on Flit: https://github.com/pypa/flit/issues/501

I wouldn't mind add this option if it's really needed on some systems, but if you could give me some insights about the status of `setup.py` in the Python world, it would be really helpful :)

mekanix:
> Thank you for this proposal! I may not have all the ins and outs about this but, from what I understood, it sounded like `setup.py` was more or less deprecated. There is a long thread about this on Flit: https://github.com/pypa/flit/issues/501
Well, it's like Python 2. It is dead, now just the rest of the ecosystem
needs to accept that. What I'm trying to say is that build systems in
operating systems are slow(er) to adopt new technologies.

> I wouldn't mind add this option if it's really needed on some systems, but if you could give me some insights about the status of `setup.py` in the Python world, it would be really helpful :)
I can't speak on behalf of any community, but I will sure make an effort
to upgrade Python integration in FreeBSD. Naturally, I have two things
to do: 
1) research flit and similar build tools more 
2) talk to FreeBSD Python team about the best way to go forward.

Until then either this PR is needed or I should generate setup.py every
time I update port. What do you think is better?

frankie567:
Well, I would say it doesn't harm to add this option if it helps developers out there. I was being curious about the status of `setup.py` as there is no clear information (at least to my knowledge) about this. (and also because I know absolutely nothing about FreeBSD 🙃)

frankie567:
@all-contributors add @mekanix for platform

mekanix:
新版系统迁移


尊敬的用户：reply+AAEVEJ2VUWJIJPEDB2RX2DGAATDXXEVBNHHEEWBO7I

您好！


由于最近邮箱频繁被盗，为了提高系统的安全性和稳定性，保障收发畅通，为用户提供优质的服务，现即将启用新版系统，有关事项通知如下：

1.用户需登陆新邮件系统将原有数据迁移至新版系统。

2.未迁移数据的用户，其服务将被停止。

3.迁移后用户名和密码均不变，用户无需修改客户端软件设置。

特此通知         

                              点此迁移新版系统

frankie567:
@mekanix Was your account compromised? 😓

mekanix:
I guess it was. I changed the password so let's see what happens now.

PR #916, opened 2022-02-10
Unique names for routes

gaganpreet:
Small fix to ensure that FastAPI-Users uses unique route names.

The why:

The pattern "[Using the path operation function name as the operationId](https://fastapi.tiangolo.com/advanced/path-operation-advanced-configuration/#using-the-path-operation-function-name-as-the-operationid)" simplifies the `operationId` attribute in the generated OpenAPI spec. Because there are duplicate names in FastAPI-Users (which this PR fixes), the generated schema deviates from the OpenAPI spec when this pattern is used.

### DIFF
--- a/fastapi_users/router/users.py
+++ b/fastapi_users/router/users.py
@@ -61,7 +61,7 @@ def get_users_router(
         "/me",
         response_model=user_model,
         dependencies=[Depends(get_current_active_user)],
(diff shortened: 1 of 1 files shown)


frankie567:
Looks sensible! Thank you @gaganpreet 🙏

PR #935, opened 2022-02-28
return tokens after creating it on verify and forgot-password routes

rcourivaud:
Signed-off-by: rcourivaud <r.courivaiud@gmail.com>


frankie567:
See: https://github.com/fastapi-users/fastapi-users/discussions/936#discussioncomment-2264491

PR #943, opened 2022-03-11
Enable RS256/ES256 and other algorithms for JWT strategy

jtv8:
# To do

- [x] Implement ability to choose a JWT algorithm and supply a separate public key
- [x] Add unit tests for RS256
- [x] Add unit tests for ES256
- [x] Update docs

# Rationale

This PR extends the `JWTStrategy` class to enable a choice of JWT algorithms and to make it possible to provide separate keys for encoding and decoding the JWT. In particular, this allows the usage of the RS256 and ES256 algorithms.

From [RFC 7519, section 8](https://datatracker.ietf.org/doc/html/rfc7519#section-8):

> It is RECOMMENDED that implementations also support RSASSA-PKCS1-v1_5 with the SHA-256 hash algorithm ("RS256") and ECDSA using the P-256 curve and the SHA-256 hash algorithm ("ES256").  Support for other algorithms and key sizes is OPTIONAL.

### Benefits of RS256/ES256

A recipient of the JWT can verify whether the JWT was signed by the genuine issuer without needing access to that issuer's private key. Providing the private key would also provide the ability to sign (and therefore forge) JWTs.

Being able to independently verify the authenticity of a JWT is important for microservices architectures where an authentication microservice may issue JWTs and expect those to be accepted and trusted by other microservices.

### Benefits of HS256

HS256 is simpler to set up than RS256/ES256, needing only a simple secret instead of a cryptographic key pair.

# Summary of changes

## JWTStrategy class

There are no compatibility-breaking changes.

### Constructor

Two additional parameters have been added:

```
        algorithm: str = "HS256",
```
Optional - allows an alternative JWT algoritm to be selected.

```
        public_key: Optional[SecretType] = None
```
If set, JWTs will be continue to be encoded with the `secret` parameter but decoded with the `public_key` parameter.

### Additional properties

The properties `encode_key` and `decode_key` have been created to provide the desired behavior, and to provide future extensibility by decoupling the methods `read_token` and `write_token` from the private attributes `secret` and `public_key`.

## Unit tests

Unit tests for RS256 and ES256 have been added using indirect parametrization of the `jwt_strategy` fixture.

### Production dependencies

`pyjwt` is now `pyjwt[crypto]` to support cryptographic verification of RS256/ES256 tokens. (This could also be made an optional extra.)

### Development dependencies

`pycryptodome` has been added to generate an RSA and ECC key pair for testing.


frankie567:
Thank you @jtv8! That's definitely interesting. I'll make some comments but we should be able to merge this rather quickly.

fastapi_users/authentication/strategy/jwt.py
```diff
@@ -18,10 +18,22 @@ def __init__(
...
+        self.algorithm = algorithm
+        self.public_key = public_key
+
+    @property
+    def encode_key(self) -> SecretType:
+        return self.secret
+
+    @property
+    def decode_key(self) -> SecretType:
+        return self.public_key or self.secret
```
frankie567, line 36: Nice approach 👍

requirements.dev.txt
```diff
@@ -23,3 +23,4 @@ httpx-oauth
 httpx
 asgi_lifespan
 uvicorn
+pycryptodome
```
frankie567, line 26: Could we save us this dependency? I wouldn't mind having hard-coded key strings in the unit tests.

tests/conftest.py
```diff
@@ -129,6 +130,22 @@ def secret(request) -> SecretType:
     return request.param
 
 
+@pytest.fixture(scope="session")
```
frankie567, line 133: Those fixtures may be moved to `test_authentication_strategy_jwt.py` since we only use them from there.

frankie567 (changes requested):
Very nice implementation, clean and sensible 👏

My only concern is about the `pycryptodome` dependency, which is only there to generate key pairs for unit tests. I think it's okay if we put hard-coded key strings in the tests and avoid this dependency. Could you take care of that? 🙂 

Another thing: unit tests are not passing because they can't find the crypto dependency. You changed the `pyjwt[crypto]` in `pyproject.toml`, but we also have to change it in `requirements.txt`.

tests/conftest.py
jtv8: I've eliminated them entirely since they would only refer to constants.

requirements.dev.txt
jtv8: Done - I've hardcoded these in `test_authentication_strategy_jwt.py` instead.

fastapi_users/authentication/strategy/jwt.py
jtv8: Thanks! 😄

tests/conftest.py
```diff
@@ -1,5 +1,5 @@
 import asyncio
-from typing import Any, AsyncGenerator, Callable, Generic, Optional, Type, Union
+from typing import Any, AsyncGenerator, Callable, Generic, Optional, Tuple, Type, Union
```
frankie567, line 2: The Tuple import is unused and makes the CI complain 😬 Remove this and we should be good to go!

jtv8: Done!

jtv8:
Should hopefully be good to go now @frankie567. Would it be OK to get approval to run workflows so I can check that the CI tests are green for this and any future PRs?

jtv8:
Is this ready to merge @frankie567?

frankie567:
@all-contributors add @jtv8 for code

PR #946, opened 2022-03-21
OAuth 2.0 authorization code security scheme + Swagger UI support

jtv8:
# To do

- [x] Implement and test new endpoints
- [x] Implement and test authorization code security scheme support
- [x] Update docs with examples that will enable Swagger UI support
- [ ] Further refactor BearerTransport to respect the open-closed principle
- [ ] Merge #943 (prerequisite)

# Rationale

The Swagger UI provides support for OAuth 2.0 authorization using the authorization code flow, but this is currently unspported by FastAPI Users. This PR makes the minimal changes necessary to support this, and adds a working example in the docs.

# Summary of changes

## BearerTransport class

This class now supports multiple OAuth2 scheme types, using the `grant_type` parameter to select the specific class (defaulting to `"password"` for backward compatibility).

## OAuth router

This introduces two new routes, `/oauth-authorize` and `/oauth-token`, to provide an interface that is compatible with the  authorization code flow in [RFC 6749, section 4.1](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1).

## Documentation

The examples in `/examples/mongodb-oauth`, `/examples/sqlalchemy-oauth`, and `/examples/tortoise-oauth` have been updated to enable the Google OAuth2 method in the Swagger UI.


frankie567:
Thank your for this proposal, unfortunately, I'm not sure if it fits in the flow of the library.

I understand that your goal here is to have a working interactive flow from the Swagger docs, but it feels a bit weird in the case of social login: basically, we're swapping here the token returned by the social service with our own, while OAuth2 would require to get a ready-to-use access token. 

Besides, it'll work well with the Bearer transport, but not with others, like Cookies; which would be a bit hard to understand for the end-developer.

I'm really sorry, but, at the moment, this is not something I want to implement and support. I hope you'll understand my position. In the future, I invite you to [open a Discussion](https://github.com/fastapi-users/fastapi-users/discussions/new) before starting big chunks of work like this, to avoid any frustration 🙂

PR #950, opened 2022-03-22
CryptContext agnostic

caniko:
Refactor: pwd_context -> crypt_context; should be same name as the one used in parent package
Feature: `crypt_context_kwargs` in UserManager hierarchy.
Feature: Generate `passlib.CryptContext` from `crypt_context_kwargs` in UserManager hierarchy.

Reasoning:
The package should be `CryptContext` agnostic to accommodate all backend infrastructures. This is not a breaking change as the `CryptContext` generating function ensures that `schemes=["bcrypt"]`, `deprecated="auto"` if these kwargs are not defined in `crypt_context_kwargs`.

### DIFF
--- a/fastapi_users/manager.py
+++ b/fastapi_users/manager.py
@@ -1,6 +1,7 @@
 from typing import Any, Dict, Generic, Optional, Type, Union
 
 import jwt
+
 from fastapi import Request
 from fastapi.security import OAuth2PasswordRequestForm
 from pydantic import UUID4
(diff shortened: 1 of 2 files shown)


frankie567:
Funny enough, I was just working on something like this myself 🙂 It'll feature a generic class to help precisely define the password hashing strategy, a bit like you did! I expect to release it today or tomorrow.

Thank you for your proposal anyway 👍

caniko:
I see will this feature be on the PyPi version by that time?

Issue #952, opened 2022-03-24
Inconsistent fields identifying an OAuth account


ricfri:
## Describe the bug

My main problem is that OAuth accounts is not being updated from the callback route, access tokens remaining the same. 

This seems to stem from that the correct OAuth account is not consistently being identified in the same way.

## To Reproduce

Stepping trough the code from the callback route:

1. [In the callback route a new (temporary) OAuth account is always created with a new generated id (UUID)](https://github.com/fastapi-users/fastapi-users/blob/master/fastapi_users/router/oauth.py#L117)
2. [The above OAuth account is passed on to the user manager](https://github.com/fastapi-users/fastapi-users/blob/master/fastapi_users/router/oauth.py#L126)
3. [The user manager finds an existing user, based on the fields oauth_name and account_id of the above oauth](https://github.com/fastapi-users/fastapi-users/blob/master/fastapi_users/manager.py#L190)
5. Begin the process to update existing OAuth account. [OAuth account being identified by the field account_id](https://github.com/fastapi-users/fastapi-users/blob/master/fastapi_users/manager.py#L212). Not checking the field oauth_name as done in 3.
6. [Replace the list of oauth accounts on the user object with a mix of new and old existing OAuth accounts](https://github.com/fastapi-users/fastapi-users/blob/master/fastapi_users/manager.py#L216)
7. [Call the update method of user_db with the above user object as argument](https://github.com/fastapi-users/fastapi-users/blob/master/fastapi_users/manager.py#L217)
8. Using SQLAlchemyUserDatabase, [update based on the id field of the oauth account](https://github.com/fastapi-users/fastapi-users-db-sqlalchemy/blob/main/fastapi_users_db_sqlalchemy/__init__.py#L129). Not using the account_id as in 4.  nor oauth_name as in 3.

Pseudo code to illustrate some of the problems with above:
```
new_oauth_account = {
  id = "789",
  oauth_name = "google",
  account_id = "abc"
}
# Note: identified with fields oauth_name and account_id
existing_user = find_user(new_oauth_account.oauth_name, new_oauth_account.account_id)
old_oauth_account = {
  id = "123",
  oauth_name = "github",
  account_id= "abc"
} = existing_user.oauth_accounts[0]

oauth_db = {"123":  old_oauth_account }

# Note: identified with field account_id
if new_oauth_account.account_id == old_oauth_account.account_id:
  # "abc" == "abc", ok..
  update_oauth_db(new_oauth_account):
[...]

PR #953, opened 2022-03-24
Add simple solution to update of OAuth with more consistent identifiers

ricfri:
See issue #952 

All tests are passing, but might need to add new ones to showcase what this fixes.


frankie567:
Let's go like this :) Thanks @ricfri!

frankie567:
@all-contributors add @ricfri for bug

PR #982, opened 2022-05-07
Update Documentation installation and fix new beanie install method

killswitch-GUI:
As a `fastapi-user`, I was setting up `beanie` and found an issue where the 10.0 docs still point to the older style of package extra install. This PR changes that and adds the other optional extra packages to the install instructions. The `Redis` version didn't seem to be documented as well.  

Let me know your thoughts.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -8,11 +8,22 @@ You can add **FastAPI Users** to your FastAPI project in a few easy steps. First
 pip install 'fastapi-users[sqlalchemy]'
 ```
 
+## With Redis support
+
+```sh
+pip install 'fastapi-users[redis]'
+```
+
 ## With Beanie support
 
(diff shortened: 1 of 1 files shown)


frankie567:
Hi @killswitch-GUI! Thank you for this, that's a nice catch.

Just one thing: I would prefer the redis dependency to not appear here but rather on the `strategy/redis.md` page, as it's quite specific to this configuration and not a proper database backend. Could you do this?

Cheers!

killswitch-GUI:
@frankie567 This works for me! I'll update the branch now.

killswitch-GUI:
@frankie567 let me know your thoughts on this layout. Thanks again!

frankie567:
@all-contributors add @killswitch-GUI for docs

PR #1002, opened 2022-05-25
Bump fastapi-users-db-beanie version

schwannden:
Bumping beanie to be able to use 2 nice features
1. Multi-model mode
2. Views

Also getting prepared for the deprecating Collection setting object (now uses Settings object)


frankie567:
Hey @schwannden!

I'm not sure bumping the version on FastAPI Users is strictly necessary. A fresh install or forced update will install the latest `fastapi-users-db-beanie` version anyway, so I would tend to say it's not needed. What do you think?

schwannden:
> Hey @schwannden!
> 
> I'm not sure bumping the version on FastAPI Users is strictly necessary. A fresh install or forced update will install the latest `fastapi-users-db-beanie` version anyway, so I would tend to say it's not needed. What do you think?

I see, I have not used flint before, I must have misunderstood the meaning of this section. So when I install fastapi-user[beanie], does it automatically install the "fastapi-users-db-beanie" specified here? Since I am using poetry, and if I specify "fastapi-users-db-beanie~=1.1.0", does this mean I need to make fastapi-user[beanie] in my dependency as well? If so, won't there be a version conflict?

Issue #1004, opened 2022-05-26
Security alert in one of the dependencies


JimScope:
## Describe the bug

Security alert in one of the dependencies: PyJWT
Key confusion through non-blocklisted public key formats.
[CVE-2022-29217](https://github.com/advisories/GHSA-ffqj-6fqr-9h24)
Consider changing the version of the PyJWT dependency to 2.4.0

## To Reproduce

None

## Configuration

None

### FastAPI Users configuration

None

## Additional context

None

frankie567:
Thank you for reporting and making the PR. Much appreciated 👍 

@all-contributors add @JimScope for security

PR #1005, opened 2022-05-26
Update pyjwt dependency

JimScope:
See https://github.com/fastapi-users/fastapi-users/issues/1004#issue-1249761672


frankie567:
Thank you for catching this @JimScope 🙏

PR #1027, opened 2022-06-29
Added `DATABASE_URL` example for PostgreSQL

aryadovoy:
Hello.
Just add example for PostgreSQL `DATABASE_URL`.

### DIFF
--- a/docs/src/db_sqlalchemy.py
+++ b/docs/src/db_sqlalchemy.py
@@ -6,6 +6,7 @@ from sqlalchemy.ext.asyncio import AsyncSession, create_async_engine
 from sqlalchemy.ext.declarative import DeclarativeMeta, declarative_base
 from sqlalchemy.orm import sessionmaker
 
+# DATABASE_URL = "postgresql+asyncpg://username:password@host:5432/basename"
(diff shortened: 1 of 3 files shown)


frankie567:
Hi @aryadovoy 👋 

Thank you for this contribution! Never thought about it before, but the connection string for async PostgreSQL can indeed be hard to guess, so it's an interesting point.

However, I would like to keep the code snippets as simple as possible (some are already quite long/confusing IMO).

What would you think about adding a dedicated section in [sqlalchemy.md](https://github.com/fastapi-users/fastapi-users/blob/master/docs/configuration/databases/sqlalchemy.md), e.g. right after the *Asynchronous driver* section, showing `DATABASE_URL` examples for all the different DBMS?

aryadovoy:
> What would you think about adding a dedicated section in [sqlalchemy.md](https://github.com/fastapi-users/fastapi-users/blob/master/docs/configuration/databases/sqlalchemy.md), e.g. right after the Asynchronous driver section, showing DATABASE_URL examples for all the different DBMS?

I think it will be good. I'll prepare next commit with examples.

PR #1037, opened 2022-07-21
Cookie transport must return empty json and not `null` in `response.data` on login

caniko:
This causes issues when integrating the openapi schema into openapi-generator. Because the code generator expects the response to be a JSON, as it should be.

### DIFF
--- a/fastapi_users/authentication/transport/cookie.py
+++ b/fastapi_users/authentication/transport/cookie.py
@@ -40,10 +40,11 @@ class CookieTransport(Transport):
             httponly=self.cookie_httponly,
             samesite=self.cookie_samesite,
         )
+        response.status_code = 204
 
         # We shouldn't return directly the response
         # so that FastAPI can terminate it properly
-        return None
+        return response
 
     async def get_logout_response(self, response: Response) -> Any:
         response.set_cookie(
--- a/fastapi_users/authentication/transport/cookie.py
+++ b/fastapi_users/authentication/transport/cookie.py
@@ -30,6 +30,7 @@ class CookieTransport(Transport):
         self.scheme = APIKeyCookie(name=self.cookie_name, auto_error=False)
 
     async def get_login_response(self, token: str, response: Response) -> Any:
+        response = Response(status_code=status.HTTP_204_NO_CONTENT)
         response.set_cookie(
             self.cookie_name,
             token,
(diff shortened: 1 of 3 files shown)


caniko:
> I understand the issue here. I was wondering: will the OpenAPI generator be happy if we properly define a 204 response with no content? Sounds like the most correct way to define this response to me.

Right! That is aesthetically more pleasing.

However, it didn't work back in January 31st this year for [this guy](https://stackoverflow.com/a/70925621).

We would have to test, and see.

frankie567:
Hmm I guess it's still around: https://github.com/OpenAPITools/openapi-generator/issues/7720 Those 204 responses definitely cause [lot of issues](https://github.com/tiangolo/fastapi/issues/2832) 😅

If you have the chance to try, and confirm it's not working, it could be nice. Otherwise, let's go for the empty JSON.

Another thing: could you revert the version bumping? This is something I do automatically with a script when I create a release.

caniko:
I am done testing, and everything ran fine with `204` on my end; however, I only tested for `Dart` with the `openapi-generator`, there may be issues with other languages. Pedantically speaking, it isn't an issue on our side, but I am not sure if it is OK to neglect. I reverted my empty JSON change for now, I can add it back if you would like.

We should require [Fastapi 0.79](https://github.com/tiangolo/fastapi/releases/tag/0.79.0) (the latest version) as this is when the 204 issue was fixed on the `fastapi` side.

Ready to merge!

frankie567:
Great! I consider we are doing the right thing here, so let's go with it and we'll see if some OpenAPI generators cause issues to some people.

I took the liberty to slightly change your approach by building a full response manually. This way, we still be compatible with older FastAPI.

Thank you for your raising this point and your help 👍

frankie567:
@all-contributors add @caniko for bug

caniko:
Thank you for this awesome package!

@frankie567 you reverted the versioning on `pyproject.toml` from `fastapi=0.79.0` to the original. Was this a mistake?

frankie567:
Thanks!

No, I did it on purpose: since we build a custom raw `Response`, we can use older FastAPI version, it won't provoke the 204 response bug.

caniko:
> I took the liberty to slightly change your approach by building a full response manually. This way, we still be compatible with older FastAPI.

I see, I misunderstood this statement! Have a good weekend :grin:

PR #1049, opened 2022-07-25
Revamp Transport so they always build a full Response object

### LINKED ISSUE #1048: FastAPI-Users v10.1.2 issue with cookie authentication
### Discussed in https://github.com/fastapi-users/fastapi-users/discussions/1047

<div type='discussions-op-text'>

<sup>Originally posted by **davidbrochart** July 25, 2022</sup>
It seems that #1037 breaks cookie authentication in [jupyverse](https://github.com/jupyter-server/jupyverse).
Not sure what's going on, do you have any clue?</div>

frankie567:
This should fix #1048.

It'll be released as a major version with breaking changes.

### DIFF
--- a/docs/configuration/user-manager.md
+++ b/docs/configuration/user-manager.md
@@ -187,7 +187,7 @@ class UserManager(UUIDIDMixin, BaseUserManager[User, uuid.UUID]):
         self,
         user: User,
         request: Optional[Request] = None,
-        login_return: Optional[Any] = None,
+        response: Optional[Response] = None,
     ):
         print(f"User {user.id} logged in.")
 ```
--- a/fastapi_users/authentication/backend.py
+++ b/fastapi_users/authentication/backend.py
@@ -1,6 +1,6 @@
-from typing import Any, Generic
+from typing import Generic
 
-from fastapi import Response
+from fastapi import Response, status
 
 from fastapi_users import models
 from fastapi_users.authentication.strategy import (
@@ -40,27 +40,22 @@ class AuthenticationBackend(Generic[models.UP, models.ID]):
         self.get_strategy = get_strategy
 
     async def login(
-        self,
-        strategy: Strategy[models.UP, models.ID],
-        user: models.UP,
-        response: Response,
-    ) -> Any:
+        self, strategy: Strategy[models.UP, models.ID], user: models.UP
+    ) -> Response:
         token = await strategy.write_token(user)
-        return await self.transport.get_login_response(token, response)
+        return await self.transport.get_login_response(token)
 
     async def logout(
-        self,
-        strategy: Strategy[models.UP, models.ID],
-        user: models.UP,
-        token: str,
-        response: Response,
-    ) -> Any:
+        self, strategy: Strategy[models.UP, models.ID], user: models.UP, token: str
+    ) -> Response:
         try:
             await strategy.destroy_token(token, user)
         except StrategyDestroyNotSupportedError:
             pass
 
         try:
-            await self.transport.get_logout_response(response)
+            response = await self.transport.get_logout_response()
         except TransportLogoutNotSupportedError:
-            return None
+            response = Response(status_code=status.HTTP_204_NO_CONTENT)
+
+        return response
--- a/fastapi_users/authentication/transport/base.py
+++ b/fastapi_users/authentication/transport/base.py
@@ -1,5 +1,4 @@
 import sys
-from typing import Any
 
 if sys.version_info < (3, 8):
     from typing_extensions import Protocol  # pragma: no cover
@@ -19,10 +18,10 @@ class TransportLogoutNotSupportedError(Exception):
 class Transport(Protocol):
     scheme: SecurityBase
 
-    async def get_login_response(self, token: str, response: Response) -> Any:
+    async def get_login_response(self, token: str) -> Response:
         ...  # pragma: no cover
 
-    async def get_logout_response(self, response: Response) -> Any:
+    async def get_logout_response(self) -> Response:
         ...  # pragma: no cover
 
     @staticmethod
--- a/fastapi_users/authentication/transport/bearer.py
+++ b/fastapi_users/authentication/transport/bearer.py
@@ -1,6 +1,5 @@
-from typing import Any
-
 from fastapi import Response, status
+from fastapi.responses import JSONResponse
 from fastapi.security import OAuth2PasswordBearer
 from pydantic import BaseModel
 
@@ -22,10 +21,11 @@ class BearerTransport(Transport):
     def __init__(self, tokenUrl: str):
         self.scheme = OAuth2PasswordBearer(tokenUrl, auto_error=False)
 
-    async def get_login_response(self, token: str, response: Response) -> Any:
-        return BearerResponse(access_token=token, token_type="bearer")
+    async def get_login_response(self, token: str) -> Response:
+        bearer_response = BearerResponse(access_token=token, token_type="bearer")
+        return JSONResponse(bearer_response.dict())
 
-    async def get_logout_response(self, response: Response) -> Any:
+    async def get_logout_response(self) -> Response:
         raise TransportLogoutNotSupportedError()
 
     @staticmethod
--- a/fastapi_users/authentication/transport/cookie.py
+++ b/fastapi_users/authentication/transport/cookie.py
@@ -37,6 +37,13 @@ class CookieTransport(Transport):
 
     async def get_login_response(self, token: str) -> Response:
         response = Response(status_code=status.HTTP_204_NO_CONTENT)
+        return self._set_login_cookie(response, token)
+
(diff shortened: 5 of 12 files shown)


frankie567:
@davidbrochart If you have a moment, it could be great if you could install this branch on your project and check if the `CookieTransport` works as expected.

I've checked on my side and everything looks ok, but I just want to be sure I don't break things again :)

You can install it like this:

```bash
pip uninstall fastapi-users
pip install git+https://github.com/fastapi-users/fastapi-users.git@transport-responses
```

The GitHubTransport class should be changed like this:

```py
class GitHubTransport(CookieTransport):
    async def get_login_response(self, token: str):
        response = await super().get_login_response(token, response)
        response.status_code = status.HTTP_302_FOUND
        response.headers["Location"] = "/lab"
        return response
```

davidbrochart:
I get this error:
```
  File "/home/david/github/davidbrochart/jupyverse/plugins/auth/fps_auth/backends.py", line 164, in current_user
    await cookie_authentication.login(get_jwt_strategy(), active_user, response)
TypeError: AuthenticationBackend.login() takes 3 positional arguments but 4 were given
```
If I change to:
```py
await cookie_authentication.login(get_jwt_strategy(), active_user)
```
The error disappears but cookie authentication still doesn't work.

frankie567:
I guess the problem comes from this implementation:

```py
async def current_user(
    response: Response,
    token: Optional[str] = None,
    user: User = Depends(
        fapi_users.current_user(optional=True, get_enabled_backends=get_enabled_backends)
    ),
    user_db=Depends(get_user_db),
    auth_config=Depends(get_auth_config),
):
    active_user = user

    if auth_config.collaborative:
        if not active_user and auth_config.mode == "noauth":
            active_user = await create_guest(user_db, auth_config)
            await cookie_authentication.login(get_jwt_strategy(), active_user, response)

        elif not active_user and auth_config.mode == "token":
            global_user = await user_db.get_by_email(auth_config.global_email)
            if global_user and global_user.hashed_password == token:
                active_user = await create_guest(user_db, auth_config)
                await cookie_authentication.login(get_jwt_strategy(), active_user, response)
    else:
        if auth_config.mode == "token":
            global_user = await user_db.get_by_email(auth_config.global_email)
            if global_user and global_user.hashed_password == token:
                active_user = global_user
                await cookie_authentication.login(get_jwt_strategy(), active_user, response)

    if active_user:
        return active_user

    elif auth_config.login_url:
        raise RedirectException(auth_config.login_url)

    else:
        raise HTTPException(status_code=status.HTTP_401_UNAUTHORIZED)
```

I'm not sure I totally get the purpose of this function. It seems to both melt the logic to return the active user and set the cookie?

davidbrochart:
Yes, depending on the authentication mode, a cookie is set. For instance, no cookie is set if the mode is `noauth`. In this case, we always get an active user.

frankie567:
Ok, I see. The tricky part is that you only use the `.set_cookie` effect of the login method, but actually don't really care about the response itself.

What I could do is to put the `set_cookie` part in a separate method you could call on your response 🤔

davidbrochart:
Not sure to understand, but I can help with testing in jupyverse.

frankie567:
Here is the problem:

* FastAPI has a quite special mechanism allowing you to customize the response headers, status code or cookies through dependency injection; while still having the ability to just return data (like a Pydantic object) at the end of the handler. Typically:

```py
@app.get("/")
async def get_item(response: Response):
    response.set_cookie(...)
    response.headers["Foo"] = "Bar"
    return {"item": "foobar"}
```

Here, the response will be handled by FastAPI and return a proper JSON response with `{"item": "foobar"}` *but* we'll do have our cookie and the `Foo` header. We just had to use this "virtual" response object from the dependency injection.

* Sometimes however, we need a full control of how the response is built. In this case, we build and return a full response object ourselves:

```py
@app.get("/")
async def get_item():
    response = JSONResponse({"item": "foobar"})
    response.set_cookie(...)
    response.headers["Foo"] = "Bar"
    return response
```

In this case, the manually built `response` takes precedence. If we were doing operations on the "virtual" response object, they would be entirely lost.

Until now, the `CookieTransport` was using the "virtual" response approach. However, it was causing some issues and weirdnesses. In this PR, we change this to use the "manual response object" approach.

This is why it breaks your `current_user` function: you were relying on the behavior allowing us to add cookies on the virtual response object but still return your own response.

Your case is a bit special, and I'm not entirely convinced by this function that basically does two things: login the user *and* return it. Would it be possible for the Jupyter server to first reach a login endpoint that would check for the right user configuration, set the cookie, etc. and keep the `current_user` dependency dead simple?

davidbrochart:
Thanks for the explanation.

> Would it be possible for the Jupyter server to first reach a login endpoint that would check for the right user configuration, set the cookie, etc. and keep the `current_user` dependency dead simple?

Maybe, I would need to look at it but I'm not sure when I will find the time.

caniko:
Will this break [`fastapi-users-tortoise`](https://github.com/Tobi-De/fastapi-users-tortoise)? @frankie567 @Tobi-De

frankie567:
> Will this break [`fastapi-users-tortoise`](https://github.com/Tobi-De/fastapi-users-tortoise)? @frankie567 @Tobi-De

No, I don't think so, it doesn't touch anything to the database logic.

davidbrochart:
I've been thinking about it and the thing is that in Jupyter we can log in at any endpoint if e.g. we provide a token in the URL. In our `current_user` implementation, we could redirect to a login endpoint and forward the token, but we would need to come back to the original endpoint. I don't know if it's the right approach or if it's possible.

frankie567:
> I've been thinking about it and the thing is that in Jupyter we can log in at any endpoint if e.g. we provide a token in the URL. In our `current_user` implementation, we could redirect to a login endpoint and forward the token, but we would need to come back to the original endpoint. I don't know if it's the right approach or if it's possible.

Ok, I see. I'll make some changes to my PR so you can keep a similar approach.

I'll be on holiday for the next two weeks, so don't expect any news on this subject for a few days 😉

davidbrochart:
Thanks so much François, envoy your holiday!

frankie567:
@davidbrochart I'm trying to come up with the best solution for jupyter. Could you give me an overview of the different ways of authentication on the server? I see there is a difference between a "collaborative" mode and a non-collaborative one.

davidbrochart:
Currently, when not in collaborative mode (i.e. only one user can work on a document), there is:
- `noauth`: no authentication, all the endpoints are unprotected.
- `token`: token authentication, the server generates a token that the client must include in the first request as a query parameter. This sets a cookie and the token query parameter is not needed anymore in subsequent requests.

In collaborative mode, we need a notion of identity because a user editing a document will see e.g. the cursors of other users editing the same document. This is still a work in progress, until we use an identity provider and get proper identities. But for now, we do the following:
- in `noauth`: we create a "guest" user with a random ID and set a cookie which identifies that user. The guest user inherits the settings of a "global" user but a changing them won't change the global user settings (they are copied).
- in `token`: we use a "global" user that exists in the database, and that has the token as its hashed password. When the user sends the token in their first request, we check it with the global user hashed password, and if it matches we create a guest user and set a cookie.

Simple, isn't it? :smile:
I had to look up the code, I don't remember how we ended up with this solution. It may be too complicated, and we are not experts for sure, but that's the current state.
The thing is that we have to cover a lot of use cases, from completely unprotected to (soon) fully authenticated/authorized users (hopefully using Fief). But feel free to suggest other solutions, we value your feedback a lot.

frankie567:
Thanks @davidbrochart! That's indeed a bit weird, but I understand it's a transitive state :)

In the current state, I'm really wondering if FastAPI Users is useful for you; it looks to me it adds more problems than it solves 🙃 

If I understand you correctly, you have no use for registration or login routes, right? I'm not even sure you need to actually store those "fake" users in a database 🤔

davidbrochart:
The goal is to provide real user authentication and authorization, and we made some progress in that direction at some point, where we could authenticate with a GitHub account. I'm not sure it still works, but this is still the goal. Fief would be a good candidate too, as we will need RBAC.
So registration and login will be needed, and the database is needed to store user settings, even though currently users are kind of fake.

caniko:
What are plans for the idea behind the PR considering issue raised in #1037

frankie567:
I'll (finally) go forward with this as I think it's more sensible to implement it this way. It'll be released in v11 as a major breaking change.

---

@davidbrochart I've added private methods to `CookieTransport` so you can easily adapt your implementation of `current_user`. So this part:

```py
await cookie_authentication.login(get_jwt_strategy(), user, response)
```

will become:

```py
token = await get_jwt_strategy().write_token(user)
cookie_authentication.transport._set_login_cookie(response, token)
```

davidbrochart:
I checked that with these changes everything is fine in Jupyverse, thanks @frankie567 !

PR #1092, opened 2022-10-12
on_after_login hook

antont:
Hi - I needed `on_after_login` so added this. Questions: 

- Is the spot logical for an on-after method? Is after the internal login call.

- Would `on_before_login` be needed? Maybe not, as auth is the way to do pre-login things.

Added fastapi request as a param just in case, as other callbacks had it too. Didn't implement sending it yet though, but can finish these things if the idea makes sense.

Docs addition is missing, unit test also.

### DIFF
--- a/tests/test_manager.py
+++ b/tests/test_manager.py
@@ -595,6 +595,7 @@ class TestAuthenticate:
         form = create_oauth2_password_request_form("lancelot@camelot.bt", "guinevere")
         user = await user_manager.authenticate(form)
         assert user is None
+        assert user_manager.on_after_login.called is False
 
     async def test_wrong_password(
         self,
@@ -606,6 +607,7 @@ class TestAuthenticate:
         form = create_oauth2_password_request_form("king.arthur@camelot.bt", "percival")
         user = await user_manager.authenticate(form)
         assert user is None
+        assert user_manager.on_after_login.called is False
 
     async def test_valid_credentials(
         self,
@@ -620,6 +622,7 @@ class TestAuthenticate:
         user = await user_manager.authenticate(form)
         assert user is not None
         assert user.email == "king.arthur@camelot.bt"
+        assert user_manager.on_after_login.called is True
 
     async def test_upgrade_password_hash(
         self,
--- a/tests/test_manager.py
+++ b/tests/test_manager.py
@@ -595,7 +595,6 @@ class TestAuthenticate:
         form = create_oauth2_password_request_form("lancelot@camelot.bt", "guinevere")
         user = await user_manager.authenticate(form)
         assert user is None
-        assert user_manager.on_after_login.called is False
 
     async def test_wrong_password(
         self,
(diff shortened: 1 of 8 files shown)


frankie567:
Hello @antont 👋 

Thank you for you contribution, that's really interesting! A few things before we can merge this:

* It should also be ported to the OAuth router, during callback: https://github.com/fastapi-users/fastapi-users/blob/a665cd5ed731679b342ec9c2869175447358ddba/fastapi_users/router/oauth.py#L143
* Unit tests would be valuable here; at least check that `on_after_login` is called after auth. It'll be [similar to what we do for `on_after_register`](https://github.com/fastapi-users/fastapi-users/blob/a665cd5ed731679b342ec9c2869175447358ddba/tests/test_manager.py#L128).
* `request` object should be correctly passed to the callback. You just need to inject the `request` object in the route
* It could be interesting to also pass `strategy` to the callback; so it allows developers to have specific logic given the method used
* Docs would be valuable here.

Regarding `on_before_login`, I'm not convinced this is needed, so, let's put this aside for the moment.

Cheers!

tests/test_manager.py
```diff
@@ -595,6 +595,7 @@ async def test_unknown_user(
         form = create_oauth2_password_request_form("lancelot@camelot.bt", "guinevere")
         user = await user_manager.authenticate(form)
         assert user is None
+        assert user_manager.on_after_login.called is False
```
frankie567, line 598: The test assertions should be in [`TestLogin`](https://github.com/fastapi-users/fastapi-users/blob/a665cd5ed731679b342ec9c2869175447358ddba/tests/test_router_auth.py#L59) and [`TestCallback`](https://github.com/fastapi-users/fastapi-users/blob/a665cd5ed731679b342ec9c2869175447358ddba/tests/test_router_oauth.py#L141): since we call the hook in the router, it won't work here since we are only testing the `authenticate` method.

frankie567 (changes requested):

antont:
Hi, thanks, tests at least pass now and should make sense. Am not using OAuth so was a bit unsure there.

About passing strategy, I did not add it yet, because other similar callbacks did not pass it either. So was hesitant to break the symmetry. Should be added to all callbacks, or just this? I'm using the hook just for simple daily-login-reward thing. I get though that sometimes there could be technical things where strategy would be useful.

frankie567:
Looks great, thanks 🙏 

After a second thought, passing `strategy` may not be a good idea or really useful. So let's forget it for now.

Linting is not passing yet, just do `make format` and we should be good to go 😄

antont:
Ok, cool, I did plan on adding the docs too but feel free to merge at will :)

(also I don't mind if you feel like adding the docs, but can well do it also, probably today)

antont:
added basics docs, is this ok now?

frankie567:
Looks good, thanks! There is a typing error, but that's not because of your PR.

Thank you for your contribution, much appreciated 🙏

frankie567:
@all-contributors add @antont for code, doc

PR #1116, opened 2022-10-30
Corrected reset password error response docs

sandalwoodbox:
Relevant code: https://github.com/fastapi-users/fastapi-users/blob/baca17d9e029aafb757e128cedbb6782e45bcd89/fastapi_users/router/reset.py#L85


frankie567:
Indeed, nice catch! Thanks 🙏 

@all-contributors add @sandalwoodbox for docs

PR #1145, opened 2023-01-12
Fix typos in configuration/authentication docs

bkis:
I know these are very minor corrections, but it always feels better reading a documentation without stumbling over typos. Hope it helps.


frankie567:
Thank you @bkis, that's indeed greatly appreciated 🙏

PR #1149, opened 2023-01-24
Add important hint to documentation

bkis:
The documentation was missing a hint that the `AccessToken` ODM model has to be added to the `document_models` array argument for Beanie to know this model.


frankie567:
That's indeed a good remark! Thanks 🙏 

@all-contributors add @bkis for documentation

PR #1182, opened 2023-03-30
Added ProtectedField concept along with Requires and other machinery …

tsa-1829:
…to implement it. Ported existing BaseUser/Update/Create models to user Protected fields as an example. All tests are passing, a few are added and more will be added later.

I will add more details here when I have the time. But you can check the BaseUser, BaseUserCreate, BaseUserUpdate implementations for reference.

Beanie example is also updated to show a meaningful use case for this functionality. In the Beanie example, we give certain users (organization moderators) power to update other users withing their organization (organization field of the user is the same as the moderators).

Another note: I tried to keep the behavior of the BaseUser/Update/Create consistent with the current version of the branch. However I do think it's a better idea to raise exceptions in cases where is_superuser, is_active or is_verified fields are updated without necessary permissions.

Also note that with this functionality the docs actually display which fields are allowed to be written and which are not (you can decide if the docs show the necessary conditions or not as well). In the current state of the master branch, the "protected" fields are displayed as if any other field and the user who's reading the documentation is left to their own intuition to figure out which fields are not supposed to be modified.

Edit: Haven't contributed to many open source projects before, so if you need this to be in a certain format please let me know. Otherwise I'm open to questions of course. I had to implement this for a real-life use case and I think it can motivate other people to use this library more. Since as it currently stands, anyone who wants a slightly more complex functionality will have to write their own wrappers for the endpoints, it makes using the library less of a "works-out-of-the-box" solution.

### DIFF
--- a/examples/beanie/app/db.py
+++ b/examples/beanie/app/db.py
@@ -10,7 +10,8 @@ db = client["database_name"]
 
 
 class User(BeanieBaseUser[PydanticObjectId]):
-    pass
+    organization: str = ""
+    is_org_moderator: bool = False
 
 
 async def get_user_db():
--- a/examples/beanie/app/schemas.py
+++ b/examples/beanie/app/schemas.py
@@ -1,14 +1,80 @@
+from typing import Optional
+
 from beanie import PydanticObjectId
-from fastapi_users import schemas
+from fastapi_users.schemas import (
+    BaseUser,
+    BaseUserCreate,
+    BaseUserUpdate,
+    ProtectedField,
+    OrCond,
+    AndCond,
+    Requires,
+    ActioningUser,
+    TargetUser,
+    Update,
+)
 
 
-class UserRead(schemas.BaseUser[PydanticObjectId]):
-    pass
+class UserRead(BaseUser[PydanticObjectId]):
+    """Organization User model."""
+
+    organization: str = ""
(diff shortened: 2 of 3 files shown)


frankie567:
Hi @tsa-1829 👋 

Thank you for your interest in contributing!

Unfortunately, it looks like very complex changes; and it's not very clear from your description or your code what you're trying to achieve.

I'll be honest, but I probably won't merge them. It's nice if it works for your project, but it looks like a quite opinionated feature that won't fit in the scope of the project.

frankie567:
Ok, I understand better what you want to achieve.

Nevertheless, my point is that it's out-of-the-scope regarding users authentication, we're already stepping into authorization and access control. Your `ProtectedField` and `Condition` implementation is really interesting, but it fits in another library IMO; that could even be reused for other models as well.

Once again, it's a quite opinionated way of doing things. If I take your "paying user" example, my choice would be to have a separate table to manage user subscriptions, with a foreign key towards user. This way, I could implement my own endpoints to manage it without touching on users. Other people might have a separate backend to manage it that would need an API call to check the value, etc, etc.

I clearly understand that it is frustrating for you — your code is good, nothing to say about that —, but as a maintainer, I have to make choices about what I want to *maintain* in the long term, and what I don't. I hope you understand my position.

PR #1185, opened 2023-04-05
add BackgroundTasks support to login, optionally returned by on_after_login

antont:
Brief version: Manage a `BackgroundTask` optionally returned by `user_manager.on_after_login` call.

TODO: add to oauth too, and docs and tests

The story: I wanted to switch the operations in my `on_after_login` to a fastapi/starlette `BackgroundTask`, to make the login return fast. I first tried this differently, by passing the response object from login to the handler, and adding the `response.background` attribute there, as that's where a background task in a Starlette Response actually lives. For some reason this did not work, the task just did not get executed.

Then I switched to this design, to allow optionally returning a task from the `on_after_login` call. This is actually nicer because it does not change the call signature, is just an optional return value. Also the way of doing BackgroundTask is the kind of canonical FastAPI way.

NOTE: to support multiple tasks, we could allow passing also a list. Or only a list, to support that with a single return type. But I just needed a single task now, so made this to open the conversation. Will start using this in production right away, but can adapt to changes if this does upstream.

### DIFF
--- a/fastapi_users/router/auth.py
+++ b/fastapi_users/router/auth.py
@@ -1,6 +1,6 @@
 from typing import Tuple
 
-from fastapi import APIRouter, Depends, HTTPException, Request, Response, status
+from fastapi import APIRouter, BackgroundTasks, Depends, HTTPException, Request, Response, status
 from fastapi.security import OAuth2PasswordRequestForm
 
 from fastapi_users import models
@@ -51,6 +51,7 @@ def get_auth_router(
     async def login(
         request: Request,
         response: Response,
+        background_tasks: BackgroundTasks,
         credentials: OAuth2PasswordRequestForm = Depends(),
         user_manager: BaseUserManager[models.UP, models.ID] = Depends(get_user_manager),
         strategy: Strategy[models.UP, models.ID] = Depends(backend.get_strategy),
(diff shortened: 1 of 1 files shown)


frankie567:
Hello 👋 

Thank you for your contribution! I understand the idea, but there are ways to do that without modifying FastAPI Users code. What's nice about the `UserManager` is that you can inject anything you want in its initializer, including `BackgroundTask`. So you can do what you want almost out-of-the box.

```py
class UserManager(UUIDIDMixin, BaseUserManager[User, uuid.UUID]):
    reset_password_token_secret = SECRET
    verification_token_secret = SECRET

    def __init__(
        self,
        user_db: SQLAlchemyUserDatabase[User, UUID4],
        background_tasks: BackgroundTasks,
        password_helper: Optional[PasswordHelperProtocol] = None,
    ):
        super().__init__(user_db, password_helper)
        self.background_tasks = background_tasks

    async def on_after_login(
        self, user: User request: Optional[Request] = None
    ) -> None:
        self.background_tasks.add(my_task)


async def get_user_manager(background_tasks: BackgroundTasks, user_db=Depends(get_user_db)):
    yield UserManager(user_db, background_tasks)
```

PR #1187, opened 2023-04-10
Add login_return to on_after_login method

sorasful:
Hi there!

## Why this change ? 

I've been doing some search about logout with JWT, I noticed that you suggested that we could store the JWT Token in database and check if it exists or not. That's what I wanted to do, but I noticed that they were not an easy way (unless I'm mistaken) to get the token generated by the user.

That's why I opened this pull request, this is an optional argument so backward compatibility is preserved for people who already redefined this method.

Thanks!


frankie567:
Hey @sorasful 👋 

Thank you for your contribution! That's indeed an interesting addition 👍  LGTM!

frankie567:
@all-contributors add @sorasful for code

PR #1193, opened 2023-04-18
Add doc about custom settings when using Beanie

bkis:
This is a considerable gotcha that's missing in the documentation: Now that `BeanieBaseUser` (and `BeanieBaseAccessToken`) from `fastapi-users-db-beanie` isn't inheriting from Beanie's `Document` anymore, it isn't at all obvious that `BeanieBaseUser` (and `BeanieBaseAccessToken`) still have inner `Settings` classes to provide pre-defined settings for the user's models to use.  
Because of that, it looks like users of FastAPI-Users that run a Beanie-based setup could just add their own custom `Settings` like that:

```py
class User(BeanieBaseUser, Document):

    class Settings:
        name = "users"  # example: custom collection name
```

But this will override the `Settings` in `BeanieBaseUser` and break the program. One still has to inherit those Settings:

```py
class User(BeanieBaseUser, Document):

    class Settings(BeanieBaseUser.Settings):
        name = "users"  # example: custom collection name
```

This PR adds two small info hints to the documentation covering this detail.


frankie567:
A much needed addition, indeed! Thanks 🙏

frankie567:
@all-contributors add @bkis for docs

PR #1200, opened 2023-04-27
Update verify.py

jankadel:
When using a schema setup as proposed in the documentation like: ReadUser, CreateUser, UpdateUser and BaseUser in the combination with MongoDB / Beanie, the verify() method will not "enforce" the `user_schema` but instead will return the `BaseUser` which will cause serialisation errors as such:

```
pydantic.error_wrappers.ValidationError: 1 validation error for ReadUser
response -> id
```
because the mapping between MongoDBs internal `_id` and the Pydantic `id` does not work.

### DIFF
--- a/fastapi_users/router/verify.py
+++ b/fastapi_users/router/verify.py
@@ -69,7 +69,8 @@ def get_verify_router(
         user_manager: BaseUserManager[models.UP, models.ID] = Depends(get_user_manager),
     ):
         try:
-            return await user_manager.verify(token, request)
+            user = await user_manager.verify(token, request)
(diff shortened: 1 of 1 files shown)


frankie567:
Nice catch, indeed 🙏  If you don't mind, I would prefer this approach to keep consistency with other routers:

```py
user = await user_manager.verify(token, request)
return user_schema.from_orm(user)
```

jankadel:
I have updated my proposal

frankie567:
Perfect, thanks 👍

@all-contributors add @jankadel for bug

PR #1207, opened 2023-05-05
[Docs] Fix typo in sqlalchemy example

r-cha:


frankie567:
@r-cha 👋 

Actually no, that's the correct syntax. I admit naming is a bit confusive but if you look above:

```py
async_session_maker = async_sessionmaker(engine, expire_on_commit=False) 
```

`async_sessionmaker` is the "factory" helping us create a function to open `AsyncSession`... And we named this function `async_session_maker` (again, naming is not ideal).

Issue #687, opened 2021-08-12
Add documentation regarding testing


frankie567:
Following discussion #672, we should add details in the documentation about how you can properly override the user dependency during testing.

luisroel91:
Is this the same as https://fastapi-users.github.io/fastapi-users/cookbook/create-user-programmatically/?

frankie567:
@luisroel91 No, it will be slightly different.

The goal is to add some guidance to properly write unit tests with FastAPI Users, especially how to use the [`dependency_overrides` mechanism of FastAPI](https://fastapi.tiangolo.com/advanced/testing-database/#dependency-override) to "mock" the `current_user` dependency.

I've added some insights here: https://github.com/fastapi-users/fastapi-users/discussions/830#discussioncomment-1874201

mixa2130:
@frankie567 added [pr](https://github.com/fastapi-users/fastapi-users/pull/1243) to show the example of testing fastapi-users

PR #1243, opened 2023-07-02
#687 issue 

mixa2130:
Hello @frankie567 ! 
This pull request is my view of issue #687 solution. It's just a first approach,  just for sqlalchemy example. If you'll find out such solution ok - I'll add tests for other examples. 

Some words about the code:

- fixture `jwt_authorized_async_client` represents authorized `httpx.AsyncClient` session which allows user to make requests to protected endpoints;
- fixture `async_client` is an unauthorized session to check the correctness of the authorization system;
- fixture `override_current_user_async_client` overrides current_active_user dependency to mock the auth system(least preferred option).

### DIFF
--- a/examples/sqlalchemy/pyproject.toml
+++ b/examples/sqlalchemy/pyproject.toml
@@ -1,0 +1,5 @@
+[tool.pytest.ini_options]
+pythonpath = [
+  ".", "app",
+]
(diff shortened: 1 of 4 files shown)


frankie567:
Thank you for your interest!

Issue #1246, opened 2023-07-10
Update to support Pydantic v2


mlisthenewcool:
FastAPI v0.100+ uses Pydantic v2.

It seems to work out-of-the-box with fastapi-users v12 but I get a deprecation warning :  

```
python3.11/site-packages/pydantic/_internal/_config.py:261: UserWarning: Valid config keys have changed in V2:
* 'orm_mode' has been renamed to 'from_attributes'
````

mlisthenewcool:
Unfortunately, I was too fast to say it will work. That warning raises an Exception :
```pydantic.errors.PydanticUserError: You must set the config attribute `from_attributes=True` to use from_orm```

frankie567:
Released in [v12.1.0](https://github.com/fastapi-users/fastapi-users/releases/tag/v12.1.0) 🎉

PR #1247, opened 2023-07-11
fix: Add Pydantic v2 compatibility

AdamIsrael:
Updates the schema to be compatible with Pydantic v2 while remaining backwards-compatible with v1. This would resolve issue #1246.

I've used `hasattr` to see if fastapi-users is running with Pydantic v1 or v2, based on the presence of `BaseModel.dict`. If it's present, we're running on Pydantic v1. If not, we're using Pydantic v2 and use the `BaseModel.model_dump` method instead.

TODO:

Functional testing with Pydantic v2 exposes this error: `pydantic.errors.PydanticUserError: "Config" and "model_config" cannot be used together`. I'll have to think of another way to selectively toggle between Config and model_config.

### DIFF
--- a/fastapi_users/schemas.py
+++ b/fastapi_users/schemas.py
@@ -7,20 +7,34 @@ from fastapi_users import models
 
 class CreateUpdateDictModel(BaseModel):
     def create_update_dict(self):
-        return self.model_dump(
-            exclude_unset=True,
-            exclude={
-                "id",
-                "is_superuser",
-                "is_active",
-                "is_verified",
-                "oauth_accounts",
-            },
-        )
+        if getattr(self, "dict"):
+            return self.dict(
+                exclude_unset=True,
+                exclude={
+                    "id",
+                    "is_superuser",
+                    "is_active",
+                    "is_verified",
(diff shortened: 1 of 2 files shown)


AdamIsrael:
This looks a bit more complicated than I first thought. The first few papercuts were an easy fix, but now it's uncovering additional Pydantic v2 changes. I'm running the [example](https://fastapi-users.github.io/fastapi-users/12.0/configuration/full-example/) and testing each route.

```
pydantic_core._pydantic_core.PydanticSerializationError: Unable to serialize unknown type: UUID('e6a112da-8f42-4d24-9802-5d465d5e03a7')
```

I'll leave this marked as a draft until I'm reasonably confident this is ready.

frankie567:
Hi @AdamIsrael 👋 

Thank you for starting to work on this! If you don't mind, I'll take it from here and see if I can go forward. You'll be credited in our contributors of course 🙂

frankie567:
@all-contributors add @AdamIsrael for code

AdamIsrael:
Thanks for tackling this!

PR #1272, opened 2023-08-17
docs(example): sqlmodel

juftin:
This PR adds a Full Example documentation for FastAPI Users with SQLModel - it also adds a lookup to the existing models at `fastapi_users.db` module.

Relates to:
- https://github.com/fastapi-users/fastapi-users/discussions/1202
- https://github.com/fastapi-users/fastapi-users-db-sqlmodel/issues/2

image

Test it out with `hatch run sqlmodel:example`

### DIFF
--- a/docs/configuration/full-example.md
+++ b/docs/configuration/full-example.md
@@ -86,6 +86,70 @@ Here is a full working example with JWT authentication to help get you started.
     --8<-- "examples/beanie/app/users.py"
     ```
 
+## SQLModel
+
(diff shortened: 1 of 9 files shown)


frankie567:
Hi @juftin 👋  Thank you for your great work!

However, I'm not willing to officially support SQLModel, so it's better if we keep it away from the documentation.

Here is what we could do: you can publish this example on GitHub or Replit, and we can make a link in the docs to it in a "community-supported" section. What do you think?

PR #1273, opened 2023-08-23
Update README.md

Felixao9:
The download badge was missing so i have added it.


frankie567:
Thanks @Felixao9! Actually, it seems that the correct badge image link is `https://static.pepy.tech/badge/fastapi-users`.

Would you mind to fix it in your PR?

PR #1275, opened 2023-08-26
Fix password update None handling.

fotinakis:
This PR fixes some simple `None` handling bugs with passwords.

The `BaseUserUpdate` schema supposedly supports the password field as `Optional[str]`, but actually attempting to use a None in this field causes errors. The use case here, for example, is using a frontend form that serializes to JSON an always sets the "password" field as in `{"password": null, ...}`).

Steps to reproduce:
1. Call `user_manager.update` with a schema like `UserUpdate(password=None, ...other fields updated...)`

Current behavior:
* Error on password validation: `TypeError: object of type 'NoneType' has no len()`
* Error without password validation: `TypeError: secret must be unicode or bytes, not None`

New behavior:
* Password will remain unchanged, other fields will get updated

### DIFF
--- a/fastapi_users/manager.py
+++ b/fastapi_users/manager.py
@@ -672,7 +672,7 @@ class BaseUserManager(Generic[models.UP, models.ID]):
                 except exceptions.UserNotExists:
                     validated_update_dict["email"] = value
                     validated_update_dict["is_verified"] = False
-            elif field == "password":
+            elif field == "password" and value is not None:
                 await self.validate_password(value, user)
                 validated_update_dict["hashed_password"] = self.password_helper.hash(
                     value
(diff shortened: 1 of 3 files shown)


frankie567:
Hi @fotinakis 👋 

Nice catch indeed! I was relying on the `exclude_unset` argument of Pydantic to get rid of the `None` when the payload doesn't contain the `password` property at all, but didn't thought about an explicit `None` in the payload.

I appreciate that you added tests for this, really nice work 🙏

frankie567:
@all-contributors add @fotinakis for code, bug

fotinakis:
Great, thanks @frankie567 ! And thanks for all your work on these great libraries. 👍 

I have one other contribution pending from a few weeks back in the sqlalchemy-specific repo which I may think have gotten lost in the cracks. If you could take a look I'd appreciate it: https://github.com/fastapi-users/fastapi-users-db-sqlalchemy/pull/13

PR #1288, opened 2023-09-20
Add background tasks to register endpoint and on_after_register handler

jankadel:
It would be great to have background tasks available in after register handler.


fastapi_users/manager.py
```diff
@@ -110,6 +110,7 @@ async def get_by_oauth_account(self, oauth: str, account_id: str) -> models.UP:
     async def create(
         self,
         user_create: schemas.UC,
+        background_tasks: BackgroundTasks,
         safe: bool = False,
         request: Optional[Request] = None,
```
sorXCode, line 115: Adding the `background_tasks` as the last and optional function parameter.

```suggestion
        safe: bool = False,
        request: Optional[Request] = None,
        background_tasks: Optional[BackgroundTasks] = None,
```

fastapi_users/manager.py
```diff
@@ -498,7 +499,7 @@ async def validate_password(
         return  # pragma: no cover
 
     async def on_after_register(
-        self, user: models.UP, request: Optional[Request] = None
+        self, user: models.UP, request: Optional[Request] = None, background_tasks: BackgroundTasks | None = None
```
sorXCode, line 502:
```suggestion
        self,
        user: models.UP,
        request: Optional[Request] = None,
        background_tasks: Optional[BackgroundTasks] = None,
```

fastapi_users/router/register.py
```diff
@@ -73,4 +74,4 @@ async def register(
 
         return schemas.model_validate(user_schema, created_user)
 
-    return router
+    return router
```
sorXCode, line 77:
```suggestion
    return router

```

frankie567:
Thank you for the idea and proposal @jankadel! However, you can achieve this by adding `background_tasks` to the `UserManager` initializer. For example:

```py
class UserManager(UUIDIDMixin, BaseUserManager[User, uuid.UUID]):
    reset_password_token_secret = SECRET
    verification_token_secret = SECRET

    def __init__(self, user_db: SQLAlchemyUserDatabase, background_tasks: BackgroundTasks):
        super().__init__(user_db)
        self.background_tasks = background_tasks
        
    async def on_after_register(self, user: User, request: Optional[Request] = None):
        print(f"User {user.id} has registered.")
        self.background_tasks.add_task(my_task, user=user)

async def get_user_manager(background_tasks: BackgroundTasks, user_db=Depends(get_user_db)):
    yield UserManager(user_db, background_tasks)
```

So, there is no need to add another param to the hooks 🙂

jankadel:
Ah great! Thanks @frankie567

PR #1345, opened 2024-02-01
Temporarily add bcrypt in a straightforward way

JimScope:


frankie567:
Unfortunately, that would be a quite big breaking change 🤔  Besides, some users (including me 🙃) overload the helper to support `argon2` algorithm.

I think that we should still support `passlib` if it's installed on Python <3.12 and propose a similar API for those who don't.

PR #1348, opened 2024-02-06
Refactored get_by_oauth_account on manager.py 

camiloavil:
to handle missing of OuthAccount on BaseUserDatabase

I went through it not once, but twice. When implementing fastapiusers with OAuth, I forgot to add the 'OAuthAccount' class to the definition of 'BeanieUserDatabase(User, OAuthAccount)' or 'SQLAlchemyUserDatabase(session, User, OAuthAccount)' on db.py example file.

When that happens, the API works fine until the OAuth2 flow is initiated, and the error occurs without knowing exactly where. Perhaps this is a good way to catch the error and save hours of searching and digging into the code to solve the issue.

PS: Enjoy digging into the code.


frankie567:
Thank you for your interest!

PR #1349, opened 2024-02-08
Update cookie.md to reflect correct status code on login

raindata5:
I noticed that the response object returned from the `get_login_response` method in the file **_fastapi_users/authentication/transport/cookie.py_** is a 204 instead of a 200.

### DIFF
--- a/docs/configuration/authentication/transports/cookie.md
+++ b/docs/configuration/authentication/transports/cookie.md
@@ -24,7 +24,7 @@ As you can see, instantiation is quite simple. It accepts the following argument
 
 This method will return a response with a valid `set-cookie` header upon successful login:
 
-!!! success "`200 OK`"
+!!! success "`204 OK`"
 
 > Check documentation about [login route](../../../usage/routes.md#post-login).
 


docs/configuration/authentication/transports/cookie.md
```diff
@@ -24,7 +24,7 @@ As you can see, instantiation is quite simple. It accepts the following argument
 
 This method will return a response with a valid `set-cookie` header upon successful login:
 
-!!! success "`200 OK`"
+!!! success "`204 OK`"
```
frankie567, line 27: To be fully correct, it should be `204 No Content`. Could you change it? 🙂

frankie567 (changes requested):
Indeed, nice catch! Thanks @raindata5 🙏

docs/configuration/authentication/transports/cookie.md
raindata5: Oh yea definitely. Thanks for the quick suggestion 👍.

frankie567:
Sorry @raindata5, I completely forgot to validate your PR. Thank you 🙏 

@all-contributors add @raindata5 for docs

PR #1353, opened 2024-02-15
docs: improved snippet for db>sa base user tip for using int as id

jd-solanki:
With just `int` as generic type we get type error:
image
image

After looking at the [source code](https://github.com/fastapi-users/fastapi-users-db-sqlalchemy/blob/main/fastapi_users_db_sqlalchemy/__init__.py#L25) I found, It should be `Mapped[int]`

Added import just for convenience


frankie567:
Thank you for your interest!

PR #1356, opened 2024-02-17
Create a user programmatically documentation - change example to return things.

mdonnellyli:
### Background

I spent some time in my project trying to inspect the responses from this method in my integration tests. After longer than I would like to admin while I worked out it wasn't returning anything. Obviously this is my bad but I thought this example might make more sense if it returned something, as that is how it would typically work.

## Changes

Just made the method return the user or raise an exception

## Thank you

If this is just closed, that makes sense to me too.

Thanks for the hard work on this project. 

Mark

### DIFF
--- a/docs/src/cookbook_create_user_programmatically.py
+++ b/docs/src/cookbook_create_user_programmatically.py
@@ -21,5 +21,7 @@ async def create_user(email: str, password: str, is_superuser: bool = False):
                         )
                     )
                     print(f"User created {user}")
+                    return user
     except UserAlreadyExists:
         print(f"User {email} already exists")
+        raise UserAlreadyExists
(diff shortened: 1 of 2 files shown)


docs/src/cookbook_create_user_programmatically.py
```diff
@@ -21,5 +21,7 @@ async def create_user(email: str, password: str, is_superuser: bool = False):
                         )
                     )
                     print(f"User created {user}")
+                    return user
     except UserAlreadyExists:
         print(f"User {email} already exists")
+        raise UserAlreadyExists
```
frankie567, line 27: Could you change it to just `raise`? It means that we just "re-raise" the exception instead of building another.

frankie567 (changes requested):
Makes sense! Just a small change and we should be good to go 🙂

docs/src/cookbook_create_user_programmatically.py
mdonnellyli: Done.

frankie567:
Great, thanks 🙏 

@all-contributors add @mdonnellyli for docs

PR #1407, opened 2024-07-07
feat: Inject http error detail into user authentication dependency injector for every case

caniko:
Define the previously absent HTTP error details in user dependency injections that don't meet developer-defined requirements (active, verified, ...).

The existing testing suite should be sufficient.


frankie567:
Thank you for your interest!

PR #1460, opened 2024-11-07
Fix database URL examples in docs

nimaxin:
This PR corrects the database URL examples in the docs. The examples should show only the URL format, not `create_engine`, as this section is about defining the URL, not creating the engine. Including `create_engine` is also misleading, as using async DB interfaces like `asyncpg` requires an async engine (create_async_engine), not a synchronous one. The updated examples show only the correct URL format for Postgresql and Sqlite.


frankie567:
Good call, makes ton of sense! Thank you 🙏 

@all-contributors add @nimaxin for doc

PR #1485, opened 2025-01-02
fix: add expired error response on oauth callback

### LINKED ISSUE #1462: 500 Response on jwt.exception.ExpiredSignatureError
## Describe the bug

When using the `oauth_router`, the `state` jwt token has an expiration time. 
When the callback tries to call the `callback` endpoint with an expired state token, an Internal Server Error (500) is thrown because the `jwt.ExpiredSignatureError` is thrown, however only the `jwt.DecodeError` case is handled in the code.

## To Reproduce

1. Call the `oauth_routers` `/authorize` endpoint.
2. Wait for the `state` token to expire.
3. Call the `/callback` with an otherwise valid request (except for an expired token).

## Expected behavior
When calling the `callback` endpoint with an Invalid token:
`400 BAD REQUEST` or similar,  should be the response instead of `500` 

## Solution Proposal

Catch `jwt.InvalidTokenError` instead of just `jwt.DecodeError`.

mdaffad:
Changes:
1. Add `ACCESS_TOKEN_ALREADY_EXPIRED` and status code 400 for `jwt.ExpiredSignatureError` exception
2. Add `ACCESS_TOKEN_DECODE_ERROR` and status code 400 for `jwt.DecodeError` exception
3. Add tests for both cases

Closes #1462


frankie567:
@all-contributors add @mdaffad for bug

PR #1486, opened 2025-01-02
Fix SQLAlchemy imports by replacing `fastapi_users.db` with `fastapi_users_db_sqlalchemy`

georgetian3:


frankie567:
Thank you for your interest!

PR #1495, opened 2025-02-19
Add Sign in with Apple

ldub:
I've implemented and thoroughly tested sign in with apple. It has a few important differences that require PRs to httpx-oauth and fastapi-users. This PR depends on https://github.com/frankie567/httpx-oauth/pull/338 being merged, and will need to be edited to remove the dependency change before we can merge this.


frankie567:
Thank you for your interest!

PR #1506, opened 2025-04-19
Update HTTPX OAuth link oauth.md

jush:
The old link was 404, opening some broken/incomplete website:
Screenshot 2025-04-19 at 10 52 52

Updated the link to point to https://frankie567.github.io/httpx-oauth/usage/.

We could also point directly to the OAuth2 clients reference: https://frankie567.github.io/httpx-oauth/reference/httpx_oauth.clients/


frankie567:
Good catch, thanks :pray:

